Issue #15, opened 2016-01-25
Ubuntu build error: format ‘%d’ expects argument of type ‘int’


pannous:
```
mpic++ -c Source/CNTK/ModelEditLanguage.cpp -o .build/Source/CNTK/ModelEditLanguage.o -D_POSIX_SOURCE -D_XOPEN_SOURCE=600 -D__USE_XOPEN2K -DCPUONLY -DUSE_ACML -DNDEBUG -msse3 -std=c++0x -std=c++11 -fopenmp -fpermissive -fPIC -Werror -fcheck-new -Wno-error=literal-suffix -O4 -ISource/Common/Include -ISource/Math -ISource/CNTK -ISource/ActionsLib -ISource/ComputationNetworkLib -ISource/SGDLib -ISource/SequenceTrainingLib -ISource/CNTK/BrainScript -I/opt/acml5.3.1/open64_64//include -I/opt/opencv-3.0.0/include -MD -MP -MF .build/Source/CNTK/ModelEditLanguage.d

Source/Math/ConvolutionEngine.cpp: In instantiation of ‘static std::unique_ptr<Microsoft::MSR::CNTK::ConvolutionEngineFactory<ElemType> > Microsoft::MSR::CNTK::ConvolutionEngineFactory<ElemType>::Create(int, Microsoft::MSR::CNTK::ConvolutionEngineFactory<ElemType>::EngineType, Microsoft::MSR::CNTK::ImageLayoutKind) [with ElemType = float]’:
Source/Math/ConvolutionEngine.cpp:493:16:   required from here
Source/Math/ConvolutionEngine.cpp:490:17: error: format ‘%d’ expects argument of type ‘int’, but argument 2 has type ‘Microsoft::MSR::CNTK::ConvolutionEngineFactory<float>::EngineType’ [-Werror=format=]
     RuntimeError("Not supported convolution engine type: %d.", engType);
  cc1plus: all warnings being treated as errors

> dpkg --list | grep compiler
ii  gcc-4.8                                                     4.8.5-1ubuntu1                             amd64        GNU C compiler
ii  gcc-4.9                                                     4.9.3-5ubuntu1                             amd64        GNU C compiler
ii  gcc-5                                                       5.2.1-22ubuntu2                            amd64        GNU C compiler
```

PR #16, opened 2016-01-25
Fixed Source/Math/ConvolutionEngine.cpp error: format ‘%d’ expects argument of type ‘int’

pannous:
Fixed https://github.com/Microsoft/CNTK/issues/15


alexeyo26:
Thank you very much for this request. We're still working on final GitHub move tasks and this will affect the time to review the requests.

Thank you for the understanding and we will come back soon.

Thank you,
 Alexey

chrisbasoglu:
Thanks for you contribution!

PR #45, opened 2016-01-28
Cuda elementwise rework

uxack:


mahilleb-msft:
@uxack did you mean to pick up the work from the CUDA-elementwise-rework branch? I don't see any commits added on top of the already existing ones... I think the existing branchs needs more work, and at this stage doesn't even merge cleanly...

frankseide:
This branch has been merged a few weeks ago and is not active. But there is certainly lots we can improve w.r.t. elementwise tensor operations.

PR #51, opened 2016-01-28
Fix the compile issue for Kaldi2Reader.

yzhang87:
Simply pass an empty string to latticesource (we are not really using it, so it's OK).


dongyu888:
merged. thanks.

PR #63, opened 2016-01-30
Add experimental support for OpenBLAS library

rwightman:
* Add 'openblas' as mathlib option in configure. Not added to auto-search so
must be specified using --with-openblas

* configure script searches outside of tail so that libraries at default_path_list
roots (ie /usr/local/openblas_config.h) are found

* Treat ACML as the odd library out in ifdefs since it doesn't conform to typical
BLAS standard. Other libraries like ATLAS should be able to share
OpenBLAS/MKL variants. Add default USE_ACML define in VS projects to match

* Fix 'max' macro define colliding with C++ std::max once openblas headers are included

Usage Notes:

* For best performance, build OpenBLAS with USE_OPENMP=1. When running CNTK, define
OPENBLAS_NUM_THREADS=#physical cores in environment or performance will suffer

* OpenBLAS 2.16 (git HEAD) tested in Linux with GCC 4.8.4 and in Windows with
OpenBLAS 2.15 (pre-built binary release + MingGW 64-bit support dlls)

* For Windows, in Math.vcxproj, replace libacml_mp_dll.lib with libopenblas.dll.a and remove
USE_ACML define. Change ACML_PATH environment variable to your OpenBLAS path.
Modify openblas_config.h as per xianyi/OpenBLAS#708

* On current generation Intel processors, OpenBLAS measures a little faster than
AMD ACML and slower than Intel MKL on MNIST and other examples

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,13 @@ ifeq ("$(MATHLIB)","mkl")
   CPPFLAGS += -DUSE_MKL
 endif
 
+ifeq ("$(MATHLIB)","openblas")
+  INCLUDEPATH += $(OPENBLAS_PATH)/include
+  LIBPATH += $(OPENBLAS_PATH)/lib
+  LIBS += -lopenblas -lm -lpthread
+  CPPFLAGS += -DUSE_OPENBLAS
+endif
+
 
 ifdef KALDI_PATH
   ########## Copy includes and defines from $(KALDI_PATH)/src/kaldi.mk ##########
--- a/Source/Math/CPUMatrix.cpp
+++ b/Source/Math/CPUMatrix.cpp
@@ -25,9 +25,6 @@
 #define NOMINMAX
 #include "Windows.h"
 #else
-#ifndef max
-#define max(a, b) (((a) > (b)) ? (a) : (b))
-#endif
 #include <cfloat>
 #endif
 
@@ -38,20 +35,27 @@
 #pragma warning(disable : 4127) // conditional expression is constant; "if (sizeof(ElemType)==sizeof(float))" triggers this
 #pragma warning(disable : 4702) // unreachable code; triggered for unknown reasons
 
-#ifndef USE_MKL
-// use ACML as default.
+#ifdef USE_ACML
 // Download ACML 5.3.1 (e.g., acml5.3.1-ifort64.exe) or above
 // from http://developer.amd.com/tools/cpu-development/amd-core-math-library-acml/acml-downloads-resources/
 // Install the ifort64_mp variant (compiled with intel compiler) of the library
 // Set Environment variable ACML_PATH to C:\AMD\acml5.3.1\ifort64_mp or the folder you installed acml
 // to point to your folder for the include file and link library
 #include <acml.h> // requires ACML 5.3.1 and above
-#else
+#elif defined(USE_MKL)
 // requires MKL 10.0 and above
 #include <mkl.h>
+#else
+#ifdef _MSC_VER
+// Visual Studio doesn't define standard complex types properly
+#define HAVE_LAPACK_CONFIG_H
+#define LAPACK_COMPLEX_STRUCTURE
+#endif
+#include <cblas.h>
+#include <lapacke.h>
 #endif
 
-#ifndef USE_MKL // MKL has one additional parameter for different matrix order
+#ifdef USE_ACML // MKL has one additional parameter for different matrix order
 #define BLAS_COLMAJOR
 #else
 #define BLAS_COLMAJOR (int) MatrixOrder::ColMajor,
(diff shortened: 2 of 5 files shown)


mahilleb-msft:
Hi @rwightman, thanks for your pull request!

Support for an open-source BLAS library is definitely welcome. Can you tweak your patch a little to mark this as "Linux" and "experimental" (e.g., in the commit message, and in configure)? We don't continuously test (or even build) this, so this would give people a heads up that there might be breakages (and it's not available on Windows).
Two other minor things:
* could you separate out the parts that aren't strictly related to OpenBLAS (e.g., CUB 1.5.1?)
* Currently your patch breaks the Windows build. If you include USE_ACML in the pre-processor definitions in Math.vcxproj I think this should be fixed.

Thanks, Mark

rwightman:
@mahilleb-msft,

I started looking into the Windows build this week but realized I need to get VS 2013 re-installed. I'll get to that soon. I was planning to do the suggested re defining USE_ACML in the Math project. It also allows an easier means to do a search and replace for flipping to USE_MKL, etc.

I'll remove the CUB change. The 'max' macro change was necessary for the openblas compile as it includes a std header below the macro definition that causes a collision with std::max. 

Do you have any specific format in mind for marking Linux/experimental or is any mention of those words in the opening line fine? Will Linux still be necessary if I fix the Windows build and also verify that it works with the prebuilt Windows OpenBlas binary? http://sourceforge.net/projects/openblas/files/v0.2.15/

mahilleb-msft:
Hey @rwightman, sounds good. If you can make it work on Windows and test at least once that would be even better. For Linux, I was thinking to put in an "(experimental)" between option and default in the output of the configure script. And then, depending on whether Windows is in or not, just properly qualify it in the commit message...
Many thanks, Mark

rwightman:
Changes made, commit amended and pushed. 

I tested in Windows, works fine with the changes mentioned in commit log. Appears to perform better than ACML on a few of the examples with my Intel Haswell but not as good as MKL.

mahilleb-msft:
@rwightman, Thanks for the updates. Build and tests (without OpenBLAS) pass BTW...
Please see a couple of minor comments above. Also, due to the changes to .vcxproj files now in master, you need do merge/rebase one more time...

Thanks, Mark

rwightman:
@mahilleb-msft, 

Yes, std(m,n) - 1 is what the documentation indicates the size of superb will be. I was being defensive to avoid allocating a zero vector and since it's a temporary that's being thrown away didn't feel  std::min(std::min(m,n)-1,1) was worth the eyesore. Can be changed or commented for clarity.

There is an openblas_set_num_threads functions. In the past there have been issues with it not being supported on the Windows build and many just recommend using the environment variable but I think that's been fixed. I can try it out for consistency with the other math libs. Note that it will be the only openblas specific code that was added, the current code should work with any cblas and lapacke conforming lib without additional ifdefs.

rwightman:
@mahilleb-msft, 

Re the confusing configure hack. With the configure find function as is, it would not search for an include in just $default_path. It only searches in the combination of $default_openblas (or other library specific vars) + $default_path. 

So, with $default_paths containing /usr, /usr/local, /opt, /opt/local, and the default_openblas needing to be empty string or a specific version, it will not search for include/openblas_config.h in /usr/local. If $default_openblas is empty, it won't search anything. If $default_openblas is set to "openblas-0.2.16" it will only search for /usr/local/openblas-0.2.16/include/, /opt/openblas-0.2.16/include/, etc.

I had openblas installed direclty in /usr/local with no version sub-folder. The same held true for my opencv install. The original script would not find it because it was not located in an opencv-3.0.0 subfolder at any of the /opt /usr/local, etc locations. My change allowed opencv to also be found on my system. 

The downside to my hack is that it will locate libraries from non specified versions of the library. However, forcing the version check to be linked to having a library in a specific sub-folder is problematic. Libraries are certainly not always located in such subfolders on linux. Version checks in configure scripts often involve parsing of header files, looking for specific library filename versions, use of dpkg, etc. 

There is a better way of doing this search as a whole as the whole script is quite simple compared to most configure scripts I've encountered. I was aiming to get it to work with the fewest additional changes.

mahilleb-msft:
@rwightman. Re std::min({m,n,1}) - but then you make it too small most of the time, right? If m > 1, n > 1 the output buffer size is still 1 which is too small to call the function?

mahilleb-msft:
@rwightman. Re the configure hack: understood. I think the commit message was just off then. We should be able to keep one for loop  -- for tail in $1 '' -- with a comment before that we also check the empty tail?

rwightman:
@mahilleb-msft, right, the outer one should have been a max, so collapsing them was wrong. Hah. Will fix.

rwightman:
@mahilleb-msft, changes made and rebased against latest master

mahilleb-msft:
@rwightman: thanks again for this .. I've merged it on Saturday, as you may have noticed.

Best, Mark

PR #86, opened 2016-02-03
forgot `throw` keyword

vladon:
exception must be thrown, not just created


frankseide:
Thanks so much! I was just going to add an Issue to fix the problems you pointed out.

Note that in this case, we should call RuntimeError() instead.

frankseide:
Hi, I made the change to RuntimeError, will land soon. Thanks for this! I will close the Issue.

PR #87, opened 2016-02-03
Input(2)->HasMBLayout

vladon:
Was same condition twice: `!Input(0)->HasMBLayout() || !Input(0)->HasMBLayout()`


frankseide:
Merged manually. Thanks!!

Issue #85, opened 2016-02-03
Some post build event fails (I have the fix but cannot push into repo)


kikish:
Hi,

When compiling the CNTK, some post-build events fail because some macros relating to paths are not in quotes.

Math and MathCUDA must use the OutputPath macro in quotes.

ComputationNetworkLib, SGDLib, CNTK, EvalDll and ActionsLib must use the TargetPath macro in quotes.

ImageReader must use quotes for both source and destination directories parameters of the xcopy command.

Regards.

Arezki.

These modifications works for me but I don't have the rights to push to this repo.

Regards,

Arezki.

kikish:
Hi,

Same problem with the Math Project;

Regards.

Arezki

frankseide:
Would you mind submitting a pull request? Thanks!!

frankseide:
Or commit to a fork and point me to it? I'd love to merge these.

wolfma61:
closed - since it is marked as merged

PR #96, opened 2016-02-07
Added missing quote in case OPENCV_PATH contains spaces.

rold2007:


alexeyo26:
Thank you - we're processing the request. More soon

alexeyo26:
Attention MS Reviewer: Build 2744

mahilleb-msft:
Thank you, we will now integrate this.

rold2007:
Note that my commit is probably only a partial fix to #85 but I only corrected what was affecting my compilation.

PR #108, opened 2016-02-11
Added reinterpret_cast<long int*>(&info) to [d,s]gesvd() of CPUMatrix.cpp

odellus:
Wasn't working for me using the `gcc` version of `mpic++`, so I had to add this in.

### DIFF
--- a/Source/Math/CPUMatrix.cpp
+++ b/Source/Math/CPUMatrix.cpp
@@ -4108,7 +4108,7 @@ void CPUMatrix<ElemType>::SVD(const CPUMatrix<ElemType>& A, CPUMatrix<ElemType>&
     if (sizeof(ElemType) == sizeof(double))
     {
 #ifndef USE_MKL
-        dgesvd('A', 'A', (int) m, (int) n, reinterpret_cast<double*>(A.m_pArray), (int) lda, reinterpret_cast<double*>(SIGMA.m_pArray), reinterpret_cast<double*>(U.m_pArray), (int) ldu, reinterpret_cast<double*>(VT.m_pArray), (int) ldvt, &info);
+        dgesvd('A', 'A', (int) m, (int) n, reinterpret_cast<double*>(A.m_pArray), (int) lda, reinterpret_cast<double*>(SIGMA.m_pArray), reinterpret_cast<double*>(U.m_pArray), (int) ldu, reinterpret_cast<double*>(VT.m_pArray), (int) ldvt, reinterpret_cast<long int*>(&info));
 #else
         double wkopt;
         int lwork = -1;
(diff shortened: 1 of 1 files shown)


mahilleb-msft:
@odellus - hm this breaks all our builds (Windows / Linux; cf. below for an example), can you share what ACML / MPI and GCC version you use?
--
Source/Math/CPUMatrix.cpp:4125:269: error: cannot convert 'long int*' to 'int*' for argument '12' to 'void sgesvd(char, char, int, int, float*, int, float*, float*, int, float*, int, int*)'
         sgesvd('A', 'A', (int) m, (int) n, reinterpret_cast<float*>(A.m_pArray), (int) lda, reinterpret_cast<float*>(SIGMA.m_pArray), reinterpret_cast<float*>(U.m_pArray), (int) ldu, reinterpret_cast<float*>(VT.m_pArray), (int) ldvt, reinterpret_cast<long int*>(&info));

veikkoeeva:
Peeking at Intel documentation, both [sgesvd](https://software.intel.com/sites/products/documentation/doclib/mkl_sa/11/mkl_lapack_examples/sgesvd_ex.c.htm) and [dgesvd](https://software.intel.com/sites/products/documentation/doclib/mkl_sa/11/mkl_lapack_examples/dgesvd.htm), both of the `info` parameters seem to be of `int`.

I wonder also if the C style casts should use C++ style as per https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Pro-type-cstylecast. I'm merely an interested follower, but I think it would make sense to mention explicitly somewhere if following [C++ Core Guidelines](https://github.com/isocpp) would be preferred. I believe it would make sense, since this is an industry wide effort and it would allow one to make use of the emerging tooling. Such as [LLVM-Clang Tidy](http://clang.llvm.org/extra/clang-tidy/index.html) and [Guidelines Support Library](https://github.com/Microsoft/GSL). In addition it would set clear guidelines for casual people to try to fix even a small pieces if they see something. :) Some information [here](https://blogs.msdn.microsoft.com/vcblog/2015/12/03/c-core-guidelines-checkers-available-for-vs-2015-update-1/).

frankseide:
The specific guideline you referenced is a good one I think. If I read it right, it applies to only objects. What I would not like to see though is using C++ casts for numeric conversions, e.g. static_cast<size_t>(i) instead of (size_t) I, as that unnecessarily increases the amount of parsing a reader has to do, without benefitting type safety.

odellus:
Mark nailed it when he asked me what version of acml I was using.

In terms of the way I fixed it, `long int info;` seemed to be better given
the funky `reinterpret_cast<>` statements I threw in, but no matter.

Please disregard this PR.

On Thu, Feb 11, 2016 at 8:24 AM, Veikko Eeva <notifications@github.com>
wrote:

> Peeking at Intel documentation, both sgesvd
> <https://software.intel.com/sites/products/documentation/doclib/mkl_sa/11/mkl_lapack_examples/sgesvd_ex.c.htm>
> and dgesvd
> <https://software.intel.com/sites/products/documentation/doclib/mkl_sa/11/mkl_lapack_examples/dgesvd.htm>,
> both of the info parameters seem to be of int.
>
> I wonder also if the C style casts should use C++ style as per
> https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Pro-type-cstylecast.
> I'm merely an interested follower, but I think it would make sense to
> mention explicitly somewhere if following C++ Core Guidelines
> <https://github.com/isocpp> would be preferred. I believe it would make
> sense, since this is an industry wide effort and it would allow one to make
> use of the emerging tooling. Such as LLVM-Clang Tidy
> <http://clang.llvm.org/extra/clang-tidy/index.html> and Guidelines
> Support Library <https://github.com/Microsoft/GSL>. In addition it would
> set clear guidelines for casual people to try to fix even a small pieces if
> they see something. :) Some information here
> <https://blogs.msdn.microsoft.com/vcblog/2015/12/03/c-core-guidelines-checkers-ava%20ilable-f%20or-vs-2015-update-1/>
> .
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Microsoft/CNTK/pull/108#issuecomment-182942013>.
>



-- 
Sincerely,

Thomas O. Wood

veikkoeeva:
@frankseide A fair point there.

Issue #127, opened 2016-02-17
Some change between cuDNN 4.0 rc and cnDNN 4.0 prod will cause the build fault.


BorisJineman:
Hi Sir,

The NVIDIA just release the cuDNN 4.0 prod version. One of the function was changed in the last version.

**cudnnBatchNormalizationBackward**

In the last version of the cuDNN, NVIDIA's developer add two new parameters into this function. And that will cause our build fault. 

The currently define of this function is as bellow:
```c
/* Performs backward pass of Batch Normalization layer. Returns x gradient,
* bnScale gradient and bnBias gradient */
cudnnStatus_t CUDNNWINAPI cudnnBatchNormalizationBackward(
                                cudnnHandle_t                       handle,
                                cudnnBatchNormMode_t                mode,
                                const void                         *alphaDataDiff,
                                const void                         *betaDataDiff,
                                const void                         *alphaParamDiff,
                                const void                         *betaParamDiff,
                                const cudnnTensorDescriptor_t       xDesc, // same desc for x, dx, dy
                                const void                         *x,
                                const cudnnTensorDescriptor_t       dyDesc,
                                const void                         *dy,
                                const cudnnTensorDescriptor_t       dxDesc,
                                void                               *dx,
                                /* Shared tensor desc for the 4 tensors below */
                                const cudnnTensorDescriptor_t       dBnScaleBiasDesc,
                                const void                         *bnScale, // bnBias doesn't affect backpropagation
                                /* scale and bias diff are not backpropagated below this layer */
                                void                               *dBnScaleResult,
                                void                               *dBnBiasResult,
                                /* Same epsilon as forward pass */
                                double                              epsilon,

                                /* Optionally cached intermediate results from
                                   forward pass */
                                const void                         *savedMean,
                                const void                         *savedInvVariance );

```
Here is the release note,

> UPDATES SINCE RELEASE CANDIDATE 
> The API of cudnnBatchNormalizationBackward has been changed to include an additional set of scaling parameters (alphaParamsDiff and betaParamsDiff) applied to the dBnScaleResult and dBnBiasResult outputs of the function.
> The prior restriction of batch size 512 in all Batch Normalization routines has been removed.
> Numerical stability and performance of cudnnBatchNormalizationBackward in some cases has been improved.
> Performance of cudnnConvolutionBackwardFilter when using Algo 1 has been improved for some cases. This code path now also requires a workspace.

So here comes the question, could i just set the new parameters as "&C::One"? 

Before:
[...]

PR #130, opened 2016-02-17
Upgrade the cuDNN version from 4.0 rc to 4.0 prod

BorisJineman:
https://github.com/Microsoft/CNTK/issues/127

The NVIDIA just release the cuDNN 4.0 prod version. One of the function was changed in the last version.

**cudnnBatchNormalizationBackward**

In the last version of the cuDNN, NVIDIA's developer add two new parameters into this function. And that will cause our build fault. 

The currently define of this function is as bellow:
```c
/* Performs backward pass of Batch Normalization layer. Returns x gradient,
* bnScale gradient and bnBias gradient */
cudnnStatus_t CUDNNWINAPI cudnnBatchNormalizationBackward(
                                cudnnHandle_t                       handle,
                                cudnnBatchNormMode_t                mode,
                                const void                         *alphaDataDiff,
                                const void                         *betaDataDiff,
                                const void                         *alphaParamDiff,
                                const void                         *betaParamDiff,
                                const cudnnTensorDescriptor_t       xDesc, // same desc for x, dx, dy
                                const void                         *x,
                                const cudnnTensorDescriptor_t       dyDesc,
                                const void                         *dy,
                                const cudnnTensorDescriptor_t       dxDesc,
                                void                               *dx,
                                /* Shared tensor desc for the 4 tensors below */
                                const cudnnTensorDescriptor_t       dBnScaleBiasDesc,
                                const void                         *bnScale, // bnBias doesn't affect backpropagation
                                /* scale and bias diff are not backpropagated below this layer */
                                void                               *dBnScaleResult,
                                void                               *dBnBiasResult,
                                /* Same epsilon as forward pass */
                                double                              epsilon,

                                /* Optionally cached intermediate results from
                                   forward pass */
                                const void                         *savedMean,
                                const void                         *savedInvVariance );

```
Here is the release note,

> UPDATES SINCE RELEASE CANDIDATE 
> The API of cudnnBatchNormalizationBackward has been changed to include an additional set of scaling parameters (alphaParamsDiff and betaParamsDiff) applied to the dBnScaleResult and dBnBiasResult outputs of the function.
> The prior restriction of batch size 512 in all Batch Normalization routines has been removed.
> Numerical stability and performance of cudnnBatchNormalizationBackward in some cases has been improved.
> Performance of cudnnConvolutionBackwardFilter when using Algo 1 has been improved for some cases. This code path now also requires a workspace.

So here comes the question, could i just set the new parameters as "&C::One"? 

Before:
```C
            CUDNN_CALL(cudnnBatchNormalizationBackward(m_cudnn, mode, &C::One, &C::One, t(inT), ptr(in), t(inT), ptr(srcGrad), t(inT), ptr(grad),
                                                       t(scaleBiasT), ptr(scale), ptr(scaleGrad), ptr(biasGrad), CUDNN_BN_MIN_EPSILON, ptr(saveMean), ptr(saveInvStdDev)));
```
After:
```C
            CUDNN_CALL(cudnnBatchNormalizationBackward(m_cudnn, mode, &C::One, &C::One, &C::One, &C::One, t(inT), ptr(in), t(inT), ptr(srcGrad), t(inT), ptr(grad),
                                                       t(scaleBiasT), ptr(scale), ptr(scaleGrad), ptr(biasGrad), CUDNN_BN_MIN_EPSILON, ptr(saveMean), ptr(saveInvStdDev)));
```

### DIFF
--- a/Source/Math/CuDnnConvolutionEngine.cu
+++ b/Source/Math/CuDnnConvolutionEngine.cu
@@ -507,7 +507,7 @@ public:
         if (m_bnImpl == BatchNormImpl::CuDnn)
         {
             cudnnBatchNormMode_t mode = spatial ? CUDNN_BATCHNORM_SPATIAL : CUDNN_BATCHNORM_PER_ACTIVATION;
-            CUDNN_CALL(cudnnBatchNormalizationBackward(m_cudnn, mode, &C::One, &C::One, t(inT), ptr(in), t(inT), ptr(srcGrad), t(inT), ptr(grad),
+            CUDNN_CALL(cudnnBatchNormalizationBackward(m_cudnn, mode, &C::One, &C::One, &C::One, &C::One, t(inT), ptr(in), t(inT), ptr(srcGrad), t(inT), ptr(grad),
                                                        t(scaleBiasT), ptr(scale), ptr(scaleGrad), ptr(biasGrad), CUDNN_BN_MIN_EPSILON, ptr(saveMean), ptr(saveInvStdDev)));
         }
         else if (m_bnImpl == BatchNormImpl::Cntk)


sarvex:
Any update when this pull request will be merged. I stumbled across the same issue. It is frustrating because the developer documentation shows steps to "cudnn v4"; which yields in non compiling code always.

Alexey-Kamenev:
We will update it soon, I don't have any specific date though.

Alexey-Kamenev:
I've merged the changes into master, however, I had to change it slightly to support both versions for now.

PR #197, opened 2016-03-03
Py3K compliant TestDriver

wilrich-msft:
Based on TestDriver.py, I ran 2to3 and applied some modifications so that it runs with Python 3. Eventually, one could merge both versions and use the six module to work around incompatibilities.


mahilleb-msft:
@wilrich-msft can you remove the additional brackets introduce by your conversion?
There are some instance of "print (...)" getting converted to "print ((...))".

Issue #171, opened 2016-02-27
Problem with 02_Convolution.ndl (MNIST examples)


erymski:
If I run `02_Convolution.ndl` (as described [here](https://github.com/Microsoft/CNTK/tree/master/Examples/Image/MNIST#config-files)) then I'm getting **90%** of error:

> Final Results: Minibatch[1-625]: Samples Seen = 10000    **err: ErrorPrediction/Sample = 0.9026**    ce: CrossEntropyWithSoftmax/Sample = 2.3738461    Perplexity = 10.738614    

The wiki page states the example should give 0.87% of error.
As the same time `01_OneHidden.ndl` produces 2.3% of error, as expected.

After some research I found that the problem can be solved by changing [`learningRatesPerMB`](https://github.com/Microsoft/CNTK/blob/master/Examples/Image/MNIST/Config/02_Convolution.cntk#L45) parameter to `0.1` in `02_Convolution.ndl`. After run it produces closer error value:
> Final Results: Minibatch[1-625]: Samples Seen = 10000    err: ErrorPrediction/Sample = **0.009**    ce: CrossEntropyWithSoftmax/Sample = 0.027893296    Perplexity = 1.028286

I'm running r2016-02-08, CNTK-2016-02-08-Windows-64bit-GPU.zip.

Alexey-Kamenev:
Hello,
Just to make sure, are you running this on GPU or CPU?

erymski:
It's reproducible for both GPU and CPU.

Alexey-Kamenev:
We will update the sample soon.

PR #204, opened 2016-03-05
Fix issue https://github.com/Microsoft/CNTK/issues/171

lofrank:
Fix issue https://github.com/Microsoft/CNTK/issues/171
After change learningRatesPerMD from 0.5 to 0.05, the error rate goes to 1% (Original error rate is about 90%).


n17s:
We got it down to 0.5%
https://github.com/Microsoft/CNTK/tree/master/Examples/Image/Classification/ConvNet#convnet_mnistcntk

Issue #257, opened 2016-03-20
When we execute the plot action, It shows "The syntax of the command is incorrect."


BorisJineman:
Hi Sir,

I try to use CNTK 's plot action to render the model to an jpg file, but it always throw the "The syntax of the command is incorrect.".

I try to investigate it, and found an issue about the replace of the renderCmd,
According to the function DoTopologyPlot in OtherActions.cpp,
we want to use bellow code to replace the renderCmd, it will not replace anything,
    msra::strfun::ReplaceAll(renderCmd, wstring(L"<IN>"), outputDotFile);
    msra::strfun::ReplaceAll(renderCmd, wstring(L"<OUT>"), outputFile);
I will fixed it by bellow code,
    renderCmd = msra::strfun::ReplaceAll(renderCmd, wstring(L"<IN>"), outputDotFile);
    renderCmd = msra::strfun::ReplaceAll(renderCmd, wstring(L"<OUT>"), outputFile);

So should I submit an pull request?

Peace and Long Life,
Boris

frankseide:
Yes, please do. Thanks much!

This error is the result of a refactoring and slipped through the cracks... Thanks for finding it!

BorisJineman:
Aye Sir, I just submit an pull request, :)
https://github.com/Microsoft/CNTK/pull/258

frankseide:
Thanks so much!

mahilleb-msft:
Closed by 8e05830b81f209b69e5620d26def6096def96108 / #258.

PR #258, opened 2016-03-20
When we execute the plot action, It shows "The syntax of the command is incorrect."

BorisJineman:
When we use the 'plot' action to render model, we will get an issue "The syntax of the command is incorrect." from the 'dot.exe'.
It is caused by the rendering cmd is not correctly,

According to the function DoTopologyPlot in OtherActions.cpp,
we want to use bellow code to replace the renderCmd, it will not replace anything,
    msra::strfun::ReplaceAll(renderCmd, wstring(L"<IN>"), outputDotFile);
    msra::strfun::ReplaceAll(renderCmd, wstring(L"<OUT>"), outputFile);
I fixed it by bellow code,
    renderCmd = msra::strfun::ReplaceAll(renderCmd, wstring(L"<IN>"), outputDotFile);
    renderCmd = msra::strfun::ReplaceAll(renderCmd, wstring(L"<OUT>"), outputFile);

https://github.com/Microsoft/CNTK/issues/257


alexeyo26:
Thank you, we will now integrate it

alexeyo26:
Attention MS Team. Integration Build 3427

PR #259, opened 2016-03-21
Initial dockerfile -- cpu only

jhludwig:
This is a dockerfile for the cpu build of CNTK without 1-bit SGD.  If useful, I will be working on GPU and 1-bit SGD versions as well


wolfma61:
this is generally interesting
out supported configuration is currently Ubuntu 14.04, so this would make more sense in the context of supporting CNTK

jhludwig:
yes this builds equally well using `ubuntu:14.04` as a base image.  easy to change

PR #275, opened 2016-03-23
CPU and GPU dockerfiles

jhludwig:
here is a revision of the CPU dockerfile, now based on ubuntu:14.04.  And a first draft of a GPU dockerfile, based ultimately on ubuntu:14.04.  Both presume that the user has already downloaded the AMD ACML .tgz.   The GPU dockerfile is easiest to use with nvidia-docker, tho it runs with standard docker as well if you set all the device mappings correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,8 +123,10 @@ else
 endif
 
 ifeq ("$(MATHLIB)","acml")
-  INCLUDEPATH += $(ACML_PATH)/include
-  LIBPATH += $(ACML_PATH)/lib
+  INCLUDEPATH += $(ACML_PATH)/ifort64/include
+  INCLUDEPATH += $(ACML_PATH)/ifort64_mp/include
+  LIBPATH += $(ACML_PATH)/ifort64/lib
+  LIBPATH += $(ACML_PATH)/ifort64_mp/lib
   LIBS += -lacml_mp -liomp5 -lm -lpthread
   COMMON_FLAGS += -DUSE_ACML
 endif
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,7 @@ ifdef CUDA_PATH
 # Set up CUDA includes and libraries
   INCLUDEPATH += $(CUDA_PATH)/include
   LIBPATH += $(CUDA_PATH)/lib64
+  LIBPATH += $(CUDA_PATH)/lib64/stubs
   LIBS += -lcublas -lcudart -lcuda -lcurand -lcusparse -lnvidia-ml
 
 # Set up cuDNN if needed
--- a/configure
+++ b/configure
@@ -10,7 +10,7 @@ enable_cuda=
 
 have_acml=no
 acml_path=
-acml_check=include/acml.h
+acml_check=ifort64/include/acml.h
 
 have_mkl=no
 mkl_path=
@@ -486,7 +486,7 @@ do
                     exit 1
                 fi
             else
-                if test $(check_dir $optarg $$libzip_check) = yes
+                if test $(check_dir $optarg $libzip_check) = yes
                 then
                     libzip_path=$optarg
                 else


jhludwig:
PS if you wish to test the built versions of these, i have pushed both to dockerhub:  surround/cntk-cpu and surround/cntk-gpu

wolfma61:
thank you - we will take a look!!!!

jhludwig:
OK i forgot one note, these files do require one small change to the Makefile.  this line needs to be added:```
  LIBPATH += $(CUDA_PATH)/lib64/stubs
```
when you set up the paths for cuda.   

i have 3 makefile and config changes i want to suggest, i will post those in a separate PR

flx42:
Hey @jhludwig, nice work! I'm a maintainer of [nvidia-docker](https://github.com/nvidia/nvidia-docker) and maybe we can improve your GPU Dockerfile solution. I looked at your Dockerfile and I have multiple remarks, some of them being advanced nitpicking :). But I must admit I've never used CNTK so I might be wrong on some point.

- Why are you setting the variable SHELL?
- The list of dependencies you `apt-get` initially seems long. I think we can trim it down a bit. For instance no need for wget AND curl, no?
- Why do you need gcc-multilib/g++-multilib?
- You should clean after `apt-get update`.
- Why don't we use OpenBLAS? It will make the Dockerfile entirely isolated (no need to pre-download anything).
- Reinstalling cuDNN will waste space, we could just symlink to the existing files until the Makefile becomes smarter.
- You should use make -j"$(nproc)" when compiling.
- Most of the time it's not needed to do "make all" and then "make install", the latter is enough. OpenBLAS is an exception here.
- Retrieving zip files requires the creation of a temporary file, you can just wget and pipe the output to "tar".
- You can use the WORKDIR to git clone CNTK to the right location.
- You can use git clone --depth 1 to limit the size of the download.
- You didn't compile OpenCV in parallel, it was thus very slooooow.

And a few others, it's simpler for me to link my own version, based on your work!
https://gist.github.com/flx42/af01eef0098ecaa0c9f8
It could be a good starting point for discussion :)
Note that I haven't tested it thoroughly.

Also, I notice that you are cloning your own fork. I guess it's related to the build issues you had.
I stumbled myself into 2 build issues:
1) With OpenBLAS I had to link with liblapacke too.
2) The build system seems to have a race condition since I randomly have a gcc crash or an error message about a missing "buildinfo.h" file.

jhludwig:
Great comments, let me respond.  happy to clean it up, i just created the most expedient version that worked for me:

- why setting variable SHELL?  I always do in every Dockerfile, I recall being surprised sometime.  But not important
- wget and curl?  no prob don't need curl, that might be inherited from another Dockerfile
- gcc-multilib/g++-multilib -- i will have to review notes, but i assume one of the CNTK prereqs required, since i would never have installed otherwise
- clean after apt-get update -- agree!
- i would be happy to switch to openblas, i see that referenced in the configure app, but openblas is not currently mentioned as a setup option on the wiki setup instructions so i assumed it was not working yet.  
- reinstalling cudnn -- agree, i just did the expedient thing
- make j"$(nproc)" -- agree, was just following the CNTK install wiki blindly
- make all and make install -- i always do both, never sure what the implementors have done under the covers.
- retrieving zip files -- good idea!
- git clone depth = 1 -- excellent point!
- compile opencv in parallel -- good point!!

thanks for the gist pointer!  i will try that out tomorrow, appreciate it very much!  yes i worked from our fork so i could make changes to makefile and configure, and to control entry of the code into our project.  

appreciate the help and appreciate the work on nvidia-docker!

jhludwig:
So Felix your gist seems to work great and is cleaner than mine, i will resubmit this PR based on that.  It will help me clean up the CPU Dockerfile too.  I am very happy to move to openBLAS, as long as the CNTK team is fully supporting -- and if they are, I'd recommend updating the Wiki to reflect

flx42:
@jhludwig great to hear! Regarding the performance of OpenBLAS I don't really know how it compares with the other options, it might matter for the CPU Dockerfile. It should not for the GPU Dockerfile though because we already have cuBLAS, cuDNN...

jhludwig:
Closing this PR and will reopen with new dockerfiles based on Felix comments

PR #283, opened 2016-03-24
Makefile and configure changes for docker and for using alternative install paths

jhludwig:
I don't expect you to merge this pull request, just using it to communicate some changes that seem effective to me, for you to consider in your next linux release.  Explanation of each change follows:

Makefile:109 -- this is needed for docker image builds, as the base nvidia docker images put libcuda.so in a slightly different place than expected.

Makefile: 125-129 -- these lines permit install of the ACML package in a different location.  one var, ACML_PATH, can now be set and libs in both ifort64 adn ifort64_mp will be picked up

configure:13 -- companion change needed for the makefile change above

configure:489 -- this is just a bug fix

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,8 +123,10 @@ else
 endif
 
 ifeq ("$(MATHLIB)","acml")
-  INCLUDEPATH += $(ACML_PATH)/include
-  LIBPATH += $(ACML_PATH)/lib
+  INCLUDEPATH += $(ACML_PATH)/ifort64/include
+  INCLUDEPATH += $(ACML_PATH)/ifort64_mp/include
+  LIBPATH += $(ACML_PATH)/ifort64/lib
+  LIBPATH += $(ACML_PATH)/ifort64_mp/lib
   LIBS += -lacml_mp -liomp5 -lm -lpthread
   COMMON_FLAGS += -DUSE_ACML
 endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -106,6 +106,7 @@ ifdef CUDA_PATH
 # Set up CUDA includes and libraries
   INCLUDEPATH += $(CUDA_PATH)/include
   LIBPATH += $(CUDA_PATH)/lib64
+  LIBPATH += $(CUDA_PATH)/lib64/stubs
```
mradmila: This is needed for CPU only build, correct?

Makefile
```diff
@@ -122,8 +123,10 @@ else
 endif
 
 ifeq ("$(MATHLIB)","acml")
-  INCLUDEPATH += $(ACML_PATH)/include
-  LIBPATH += $(ACML_PATH)/lib
+  INCLUDEPATH += $(ACML_PATH)/ifort64/include
+  INCLUDEPATH += $(ACML_PATH)/ifort64_mp/include
+  LIBPATH += $(ACML_PATH)/ifort64/lib
+  LIBPATH += $(ACML_PATH)/ifort64_mp/lib
```
mradmila: I thought that ACML_PATH included ifort64_mp? No? Look here: default_acmls="acml5.3.1/ifort64_mp". I don't think ifort64 directory is necessary on the lib or include path.

wolfma61:
some of these changes have made it in - appreciate the effort!
if you have more suggestions, please open a new issue or pull request

PR #293, opened 2016-03-26
revised docker files

jhludwig:
Based on the good comments from Felix in https://github.com/Microsoft/CNTK/pull/275, here are revised docker flles -- much cleaner, switching to openBLAS from the amd math libs

### DIFF
--- a/Tools/docker/Dockerfile-CPU
+++ b/Tools/docker/Dockerfile-CPU
@@ -43,7 +43,6 @@ RUN wget -q -O - http://nih.at/libzip/libzip-1.1.2.tar.gz | tar -xzf - && \
     make -j"$(nproc)" install && \
     rm -rf /libzip-1.1.2
 
-
 RUN wget -q -O - https://github.com/Itseez/opencv/archive/3.0.0.tar.gz | tar -xzf - && \
     cd opencv-3.0.0 && \
     cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_INSTALL_PREFIX=/usr/local/opencv-3.0.0 . && \
@@ -66,7 +65,3 @@ RUN git clone --depth=1 -b master https://github.com/Microsoft/CNTK.git . && \
      make -j1 all 
 
 ENV PATH=/cntk/build/release/bin:$PATH
-
-CMD ["/bin/bash"]
-
-
--- a/Tools/docker/Dockerfile-CPU
+++ b/Tools/docker/Dockerfile-CPU
@@ -1,0 +1,72 @@
+# CNTK Dockerfile
+#   CPU only
+#   No 1-bit SGD
+#   Assumes AMD ACML has been previously downloaded and the .tgz file resides in Docker dir
+
+FROM ubuntu:14.04
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        autotools-dev \
+        build-essential \
+        cmake \
+        git \
+        g++-multilib \
+        gcc-multilib \
+        libavcodec-dev \
+        libavformat-dev \
+        libjasper-dev \
+        libjpeg-dev \
+        libpng-dev \
+        liblapacke-dev \
+        libswscale-dev \
+        libtiff-dev \
+        pkg-config \
+        python-dev \
+        python-numpy \
+        python-pip \
+        wget \
+        zlib1g-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN wget -q -O - https://www.open-mpi.org/software/ompi/v1.10/downloads/openmpi-1.10.1.tar.gz | tar -xzf - && \
+    cd openmpi-1.10.1 && \
+    ./configure --prefix=/usr/local/mpi && \
+    make -j"$(nproc)" install && \
+    rm -rf /openmpi-1.10.1
+
+ENV PATH /usr/local/mpi/bin:$PATH
+ENV LD_LIBRARY_PATH /usr/local/mpi/lib:$LD_LIBRARY_PATH
+
+RUN wget -q -O - http://nih.at/libzip/libzip-1.1.2.tar.gz | tar -xzf - && \
+    cd libzip-1.1.2 && \
+    ./configure && \
+    make -j"$(nproc)" install && \
+    rm -rf /libzip-1.1.2
+
+
+RUN wget -q -O - https://github.com/Itseez/opencv/archive/3.0.0.tar.gz | tar -xzf - && \
+    cd opencv-3.0.0 && \
+    cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_INSTALL_PREFIX=/usr/local/opencv-3.0.0 . && \
+    make -j"$(nproc)" install && \
+    rm -rf /opencv-3.0.0
+
+RUN wget -q -O - https://github.com/xianyi/OpenBLAS/archive/v0.2.17.tar.gz | tar -xzf - && \
+    cd OpenBLAS-0.2.17 && \
+    make -j"$(nproc)" && \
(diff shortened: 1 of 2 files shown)


Tools/docker/Dockerfile-GPU
```diff
@@ -0,0 +1,70 @@
...
+    ./gdk.run --silent && \
+    rm gdk.run
+
+WORKDIR /cntk
+RUN git clone --depth=1 -b master https://github.com/Microsoft/CNTK.git . && \
+    sed -i 's/-lopenblas/-lopenblas -llapacke/g' Makefile && \
+    mkdir -p build/release && \
+    cd build/release && \
+    ../../configure --with-cuda=/usr/local/cuda --with-cudnn=/usr/local/cudnn --with-openblas=/usr/local/openblas && \
+     make -j1 all # FIXME: Makefile race condition?
```
mradmila: Is there a bug on this? I can't seem to repro it.

Tools/docker/Dockerfile-GPU
```diff
@@ -0,0 +1,70 @@
...
+    mkdir -p /usr/local/cudnn/cuda/lib64 && \
+    ln -s /etc/alternatives/libcudnn_so /usr/local/cudnn/cuda/lib64/libcudnn.so
+
+RUN wget -O gdk.run -q http://developer.download.nvidia.com/compute/cuda/7.5/Prod/local_installers/cuda_352_39_gdk_linux.run && \
+    chmod +x gdk.run && \
+    ./gdk.run --silent && \
+    rm gdk.run
+
+WORKDIR /cntk
+RUN git clone --depth=1 -b master https://github.com/Microsoft/CNTK.git . && \
```
mradmila: This does not include 1-Bit SGD. Should we have a separate image for that?

Tools/docker/Dockerfile-CPU
```diff
@@ -0,0 +1,72 @@
...
+
+ENV PATH /usr/local/mpi/bin:$PATH
+ENV LD_LIBRARY_PATH /usr/local/mpi/lib:$LD_LIBRARY_PATH
+
+RUN wget -q -O - http://nih.at/libzip/libzip-1.1.2.tar.gz | tar -xzf - && \
+    cd libzip-1.1.2 && \
+    ./configure && \
+    make -j"$(nproc)" install && \
+    rm -rf /libzip-1.1.2
+
```
mradmila: Nit: extra space

Tools/docker/Dockerfile-CPU
```diff
@@ -0,0 +1,72 @@
...
+RUN git clone --depth=1 -b master https://github.com/Microsoft/CNTK.git . && \
+    sed -i 's/-lopenblas/-lopenblas -llapacke/g' Makefile && \
+    mkdir -p build/release && \
+    cd build/release && \
+    ../../configure --with-openblas=/usr/local/openblas && \
+     make -j1 all 
+
+ENV PATH=/cntk/build/release/bin:$PATH
+
+CMD ["/bin/bash"]
```
mradmila: Don't you get the same with -ti on the docker line? You don't have this in the other Dockerfile

Tools/docker/Dockerfile-GPU
```diff
@@ -0,0 +1,70 @@
...
+    make -j"$(nproc)" && \
+    make PREFIX=/usr/local/openblas install && \
+    rm -rf /OpenBLAS-0.2.17
+
+RUN mkdir -p /usr/local/cudnn/cuda/include && \
+    ln -s /usr/include/cudnn.h /usr/local/cudnn/cuda/include/cudnn.h && \
+    mkdir -p /usr/local/cudnn/cuda/lib64 && \
+    ln -s /etc/alternatives/libcudnn_so /usr/local/cudnn/cuda/lib64/libcudnn.so
+
+RUN wget -O gdk.run -q http://developer.download.nvidia.com/compute/cuda/7.5/Prod/local_installers/cuda_352_39_gdk_linux.run && \
```
mradmila: This gdk is from 7.5, which we don't support (yet). To be consistent, this should probably be "/7_0/Prod/local_installers/cuda_346.46_gdk_linux.run"

mradmila:
This looks great, thank you. I left a few comments, mostly minor. Can you please look at those and then we can do the merge?

Tools/docker/Dockerfile-GPU
flx42: I think so, I was randomly getting 2 errors on a machine with a large number of cores (56).
Either a gcc crash:
`The bug is not reproducible, so it is likely a hardware or OS problem.`
Or something like:
`buildinfo.h: No such file or directory`

It could be that one header file is included while being simultaneously generated by the Makefile. Just a guess. I will try to send a full log soon.

Tools/docker/Dockerfile-GPU
jhludwig: ah i couldn't find that file.  i will try it out.

Tools/docker/Dockerfile-CPU
jhludwig: i'll add thanks

Tools/docker/Dockerfile-GPU
jhludwig: the separate license you have for 1-bit sgd has been offputting for me.  i haven't looked at.  i guess i'd recommend that you all maintain separate dockerfiles for the separate license regimes.

Tools/docker/Dockerfile-GPU
jhludwig: yes compile files everytime for me on an 8G machine with -j1.   works ok on 16g machine

Tools/docker/Dockerfile-GPU
flx42: Good point, needs to be fixed.

Tools/docker/Dockerfile-CPU
flx42: Not exactly, `-ti` will run the default command which could be anything.
But if you don't specify any `CMD` it will default to `/bin/bash`, so yeah, it's not needed.

flx42:
@mradmila Can you confirm OpenBLAS is fine?

Tools/docker/Dockerfile-GPU
mradmila: You mean with -j alone?

Tools/docker/Dockerfile-GPU
mradmila: I completely understand, other people may feel the same way. Let's spin up a separate image for that.

Tools/docker/Dockerfile-GPU
mradmila: Actually, I think we caught this issue as well, today (I just read your comment about buildinfo.h). Just leave this as -j1 and I will change it once we fix the bug.

mradmila:
You mean, is it OK to use OpenBLAS? Absolutely. I did not, however, verify that OpenBLAS dependent workloads are working properly.

flx42:
OK! The `configure` script lists openblas as experimental, that's why I'm asking.

Another thought: maybe we can reduce the OpenCV compile time by tweaking the cmake command line. For instance, I think specifying `-DWITH_CUDA=OFF` would be fine.

jhludwig:
Re compiling opencv, in our environment we would like to have opencv compiled with cuda and are willing to incur the cost, but defer to maintainer's wishes

jhludwig:
OK i picked up the changes from above -- remove extraneous space, remove default CMD, change gdk to 7.0.

mradmila:
We would like to have OpenCV compiled with CUDA support as well, and incur the cost as necessary. The changes look good to me, let me see what procedure we need to follow for merging.

PR #304, opened 2016-03-30
Update mnist_convert_python3.py

HthSolid:
missing request specification in urllib import


mahilleb-msft:
@HthSolid : thanks for the fix. Will look at integrating it...

Best, Mark

PR #395, opened 2016-04-23
Update Network Description Language.md

polidore:
Delay seems to be deprecated in favor of PastValue and FutureValue. I found it (even more :) ) confusing reading the LSTM examples without this documentation since Delay isn't used, but PastValue is.  Also, this document had the convolution parameters for Delay, which I've also removed.


chrisbasoglu:
Thank you for your change.  Many of your edits have been incorporated into that document.

PR #397, opened 2016-04-24
FIXED: error: expected ')' before 'PRIu64' on Linux CPU-only

fabrymarko:
When compiling CPU-only flavor of CNTK on Linux it fails to build because of
```
error: expected ')' before 'PRIu64'
```
Solution, replaced all:
```
#include <stdint.h>
```
with:
```
#define __STDC_FORMAT_MACROS
#include <inttypes.h>
```

which include the #include <stdint.h> and also define PRIu64.

### DIFF
--- a/Source/Common/Include/latticearchive.h
+++ b/Source/Common/Include/latticearchive.h
@@ -17,7 +17,6 @@
 #include "latticestorage.h"
 #include "simple_checked_arrays.h"
 #include "fileutil.h"
-#include <stdint.h>
 #include <vector>
 #include <string>
 #include <unordered_map>
--- a/Source/Readers/UCIFastReader/UCIParser.h
+++ b/Source/Readers/UCIFastReader/UCIParser.h
@@ -9,7 +9,8 @@
 #include <string>
 #include <vector>
 #include <assert.h>
-#include <stdint.h>
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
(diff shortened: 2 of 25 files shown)


raaaar:
@eldakms is added to the review. #Closed

raaaar:
@mahilleb-msft is added to the review. #Closed

raaaar:

Source/Readers/UCIFastReader/UCIParser.h
```diff
@@ -9,7 +9,8 @@
 #include <string>
 #include <vector>
 #include <assert.h>
-#include <stdint.h>
+#define __STDC_FORMAT_MACROS
+#include <inttypes.h>
```
raaaar: Don't think, this replacement is really called for here (and in a few other places as well)

Source/Common/Include/latticearchive.h
```diff
@@ -17,7 +17,6 @@
 #include "latticestorage.h"
 #include "simple_checked_arrays.h"
 #include "fileutil.h"
-#include <stdint.h>
```
raaaar: Does it not compile with this left in place?

raaaar:
Prefixing all inclusions of <inttypes.h> with #define __STDC_FORMAT_MACROS should be enough to solve the issue. Removing <stdint.h> doesn't seem to be necessary.

PR #466, opened 2016-05-11
bugfix: grayscale image now assumed by default if number of channels …

DanielMerget:
Using ImageReader with grayscale images crashes CNTK unless the reader parameter "grayscale=true" is set in the cntk config. However, CNTK should be smart enough to (by default) read grayscale images if the number of channels in the feature section of the reader is one anyway (channels=1).

### DIFF
--- a/Source/Readers/ImageReader/ImageConfigHelper.cpp
+++ b/Source/Readers/ImageReader/ImageConfigHelper.cpp
@@ -77,7 +77,7 @@ ImageConfigHelper::ImageConfigHelper(const ConfigParameters& config)
 
     m_mapPath = config(L"file");
 
-    m_grayscale = config(L"grayscale", false);
+    m_grayscale = config(L"grayscale", c == 1);
     std::string rand = config(L"randomize", "auto");
 
     if (AreEqualIgnoreCase(rand, "auto"))


Alexey-Kamenev:
Hi Daniel,
Where does the crash happen? grayscale flag controls whether the image will be converted to 1 channel or 3 channel output, however, according to OpenCV documentation, `imread` should always be able to read images, both color and grayscale.

DanielMerget:
Hi Alexey,

here is the call-stack after a first-chance exception has been thrown:

first_chance

>  	opencv_world300.dll!00007ffcf7d6982c()	Unknown
>  	opencv_world300.dll!00007ffcf7d6a343()	Unknown
>  	opencv_world300.dll!00007ffcf7c8548a()	Unknown
>  	opencv_world300.dll!00007ffcf7c83948()	Unknown
> 	ImageReader.dll!Microsoft::MSR::CNTK::ImageDataDeserializer::ImageChunk::GetSequence(unsigned __int64 sequenceId, std::vector<std::shared_ptr<Microsoft::MSR::CNTK::SequenceDataBase>,std::allocator<std::shared_ptr<Microsoft::MSR::CNTK::SequenceDataBase> > > & result) Line 93	C++
>  	ImageReader.dll!Microsoft::MSR::CNTK::BlockRandomizer::GetNextSequences::__l14::<lambda>(int i) Line 131	C++
>  	ImageReader.dll!Microsoft::MSR::CNTK::BlockRandomizer::GetNextSequences$omp$1() Line 141	C++
>  	vcomp120d.dll!00007ffd21831854()	Unknown
>  	vcomp120d.dll!00007ffd2183179d()	Unknown
>  	vcomp120d.dll!00007ffd21831759()	Unknown
>  	vcomp120d.dll!00007ffd2183848c()	Unknown
>  	kernel32.dll!00007ffd27c813d2()	Unknown
>  	ntdll.dll!00007ffd2a3954e4()	Unknown

The critical lines here are in ImageDataDeserializer:
>         if (cvImage.type() != CV_MAKETYPE(dataType, cvImage.channels()))
>         {
> /*Next Statement*/    cvImage.convertTo(cvImage, dataType);
>         }

What is maybe more interesting (I only found out now that I wanted to reproduce the error) is that the crash happens only once after a full rebuild. Afterwards, everything is fine (at least on my machine). One difference I noticed was that the acquisition of the gpu is somehow delayed during the first run-through.

First run-through after rebuild (gpu is acquired after ~ a minute)


Second run-through after rebuild (gpu is acquired almost instantly)


I don't remember exactly how I found out that the grayscale boolean is the cause of this crash. Having grayscale set to false seems to resolve the issue, however, even for the first run after a rebuild (tested twice to make sure). I have not tested this for the CPU, though.

Alexey-Kamenev:
Hi Daniel,
Thanks for details reply - it helps. I asked because I could repro this crash in the unit tests but I know some of our internal users complained about the same problem as well so I'm trying to understand what might be wrong.

DanielMerget:
Hi again,

Meanwhile, I looked into something completely different, and suddenly I had a very similar crash in almost the same spot again. This new crash, however, occurs completely at random, something between 20% and 40% of the time and only in release mode. So I figured that likely some race condition might be the cause. I am currently still debugging this, trying to reproduce it with a minimal non-working example for the master branch.

My first try was to disable parallel reading of images (just comment the "#pragma omp parallel ..." line in BlockRandomizer, and just to make sure also in TransformerBase). Once I did this, the problem vanished (not reproducible even after 20+ tries). I have yet to understand why exactly it does, though, since I could not identify any clear race in the code. Maybe some opencv function used in there is not thread-safe, but that is just a guess. After all, the parallel reading might not be the actual cause.

I will look into this more a little more closely tomorrow and report back if I find anything out.

Alexey-Kamenev:
Oh, ok - I know this problem then. As you correctly pointed out, it is related to multi-threading and is happening from time to time in my (and other people) experiments as well.
To disable multi-threading you can use `numCPUThreads` option in config file (or command line) – just set it to 1. Performance will obviously suffer.
What I noticed is if crash happens then it always happens just at the very beginning. If the crash did not happen at the beginning, CNTK usually works fine for the rest of the training.
I think there might be a bug in OpenCV which causes race condition on initialization. Unfortunately, I did not have time to investigate that further.

PR #467, opened 2016-05-11
bugfix: multiview in ImageReader now disabled by default (as intended)

DanielMerget:
An empty string will (curiously) be mapped to CropType::Multiview10 instead of CropType::Center. This is #1 unintuitive and #2, for regression tasks, this is never desirable since it would result in wrongly labeled images.


Alexey-Kamenev:
I've refactored the code a bit, basically removing 2 different places where cropType was parsed. See f603104.

DanielMerget:
Thanks

PR #482, opened 2016-05-17
Sub-linear memory optimization on ResNet

yuxiaoguo:
Reference "Training Deep Nets with sublinear memory cost"
1. Add a buffer pool and re-implement GPUMatrix Resize function
2. Add ResNet-101 script on ImageNet samples.
3. Release forward data right after forwardprop and re-generate in backprop


yuxiaoguo:
@alexeyk is added to the review. #Closed

yuxiaoguo:
@alexey is added to the review. #Closed

yuxiaoguo:
@Alexey-Kamenev is added to the review. #Closed

amitaga:
Thanks for submitting the PR yuxiaoguo. This is a valuable feature and we are in the process of reviewing it.

Can you please share how much reduction in memory footprint did you observe with this optimization? Also, are your comparisons relative to "shareNodeValueMatrices=false" or "shareNodeValueMatrices=true' - this is an option that enables the optimization to reuse forward prop intermediate buffers which are not used during backprop reducing memory footprint.

Regards,
Amit

yuxiaoguo:
@amitaga 
First, about the memory reduction. From I test in local machine, in CIFAR-10 ResNet-110, before using sub-linear memory, the memory cost is 2.4G. While after enable it, the memory cost will decrease to 600M. Another experiments is ResNet-152, before importing sub-linear memory, we can't set up experiments which batch size is more than 8, now we are running that of batch size 32. Which mean we at least reduce the memory into 25%. The truth is the more sharedNode we have, the more memory we could save. For a N sharedNodes network, the algorithm will reduce the memory into Sqrt(N).

Second, I have notice the keyword shareNodeValueMatrices. Indeed, in some network, current CNTK memory manager can not finish the forward prop, like ResNet-152, which means the power of shareNodeValueMatrices is restricted. Sub-linear memory will release all the shareNode memory in forward prop, only leave some key-node to recover the forward prop info in future back prop. We add a keyword Train[SGD[enableSublinearMemory=true]] to control the feature.

However, there are some tradeoff. Although, we could release all the forward sharedNode in forward prop. We need to recover it step by step in back prop, which means another forward prop will be added. In the paper and our test, it will take extra 25% time for training.

Here is the original idea of sub-linear memory, "[Training Deep Nets with Sublinear Memory Cost](https://arxiv.org/pdf/1604.06174v2.pdf)".

Issue #422, opened 2016-05-02
ImageReader does not support regression


DanielMerget:
Hi,

**First of all some context:** I am trying to do landmark detection for faces, so I have several output labels with a SquareError loss. I got to run CNTK perfectly for this task using the UCIFastReader, but when I tried to use the ImageReader, I found that it does not support regression, i.e., it does not read more than 1 label because classification is automatically assumed in the code, the label is represented as "size_t m_classId" in ImageDataDeserializer, etc. Maybe even worse: it does so silently, so it took me a while to figure out why the regression did not work.

Landmark detection is the most likely scenario I can think of where this is an issue. But there are probably other regression tasks on images.

In the long run, it is no good option for me to use the TextReaders because I have frequently-changing datasets. Just changing the label file for the ImageReader would be much faster and also saves disk space. I need sub-pixel accuracy, so trying some discrete softmax workaround is futile.

The code which reads multiple labels for regression and the code which reads images (and crops them, etc.) is obviously all there. Still it seems like the architecture of the ImageReader differs greatly from the architecture of the UCIFastReader, for example, so bringing these parts together in a consistent and concise way is a little more involved than simple copy+paste. I know that UCIFastReader is deprecated, but I have not taken a closer look at the TextFormatReader, yet.

**The good news:** I am willing to implement the necessary changes myself and post a pull request. I already checked some of the CNTK source and got it all set up in Visual Studio. However, I am not very familiar with the project source, so I might need a few hints about where to start and how to best tackle the problem. Otherwise, I fear that it might become some messy workaround which nobody but me can use :-)

For example, would it be better to adapt the ImageReader to account for regression, or rather adapt either of the TextReaders to account for image data (or maybe inherit from it)? Which classes would be involved apart from ImageReader? Very likely ImageDataDeserializer, probably ImageConfigHelper. ImageTransformers should also have a way to apply their transformations to the labels as well (that's probably a follow-up issue). How about other classes/files? Is there maybe even some work in progress on this (I didn't find anything)?

I would appreciate any advice.

Best,
Daniel

rodrigoalor:
hello,  
coincidentally  i tried to use cntk ucifastreader  for regression for face landmarks, i try many times but   the output of the model always was the mean of the level position, in other words the model learned the mean and dont work. your model with ucifastreader works?
Could you share  the  cntk configuration or the dataset, please?
  
[...]

PR #719, opened 2016-07-27
Added support for regression in ImageDataDeserializer

DanielMerget:
**This is pull request is related to Issue #422.**

**The main contribution of this pull request:**

It is now possible to use ImageDataDeserializer (or ImageReader) to read regression labels. This works in both the "corpus" and the "old deprecated" way, namely:

* in the .cntk config file, set the (global) parameter "labelType = regression" (default is "classification"; maybe "category" could be added as a valid input as well, but I have not done so, yet)
* labelDim has to be provided: it is the length of the label vector (basically the same as for "classification")
* in the mapping file, simply write the tab-separated labels after the image path, see the two simple UnitTests I added for an example.

**An addition to the above:**

(Parts of) the labels can now also be "NAN" (or "nan" or "NaN" etc.). Any NAN labels are internally translated to std::numeric_limits<Precision>::lowest() (because it seems that cuda does not handle (Q)NAN really well). I would have loved to do it with NAN directly, but gave up after two weeks (I am not a CUDA expert though, so it may actually be possible, and it would definitely be nicer/cleaner). Anyway, these specific std::numeric_limits<Precision>::min() values are later ignored during the computation of the (square) loss and also during back-propagation. The caveat here is that this cannot be used in conjunction with any .ndl options for scaling the labels (or generally modifications), since it would quickly result in some sort of underflow. Features are not affected and can be scaled the same way as before.

This is not an essential part of "Image Regression" in general. So if that specific part raises some eyebrows or is not desirable on your part, I can remove it from the pull request. However, I find it is a useful feature for image regression, especially when trying to combine several databases (which may have different subsets of labels and thus some of them are unknown -> NAN), for example.

If you find that there are other issues or if you have questions, feel free to let me know and I can try to resolve them.

**Final remark:**

I had to merge the things manually since there were a lot of conflicts. Therefore, there is no visible history of the branch I merged with (thankfully, the 'lost' history is not too long, though).
You can find most of it here: https://github.com/DanielMerget/CNTK/tree/special_square_loss_layer
The UnitTests are from here: https://github.com/DanielMerget/CNTK/tree/image_regression

Best,
Daniel


wolfma61:
Hi Daniel - 

thank you. Let us take a look and we will get back to you. Give us a few days (next week I would guess) ... some vacationing going on in the team ;)

Thanks
Wolfgang

Source/Readers/ImageReader/ImageDataDeserializer.cpp
```diff
@@ -73,10 +65,25 @@ IDataDeserializerPtr createImageDataDeserializer(const ConfigParameters& config,
     return deserializer;
 }
 
-IDataDeserializerPtr createImageDataDeserializer(const ConfigParameters& config)
+}
+
+std::unique_ptr<IDataDeserializer> createImageDataDeserializer(CorpusDescriptorPtr corpus, const ConfigParameters& config)
+{
+    auto dbg = std::string(config(L"labelType", "classification")); UNUSED(dbg);
```
pkranen: Can those be removed?

Source/Readers/ImageReader/ImageDataDeserializer.cpp
```diff
@@ -73,10 +65,25 @@ IDataDeserializerPtr createImageDataDeserializer(const ConfigParameters& config,
...
+    return createImageDataDeserializer(corpus, config, labelType, elementType);
+}
+
+// (soon to be) deprecated version
+std::unique_ptr<IDataDeserializer> createImageDataDeserializer(const ConfigParameters& config)
 {
-    ImageConfigHelper configHelper(config);
-    return createImageDataDeserializer(config, configHelper);
+    auto dbg = std::string(config(L"labelType", "classification")); UNUSED(dbg);
+    auto dbg2 = std::string(config(L"precision", "float")); UNUSED(dbg2);
```
pkranen: Same here.

Source/Readers/ImageReader/ImageDataDeserializer.h
```diff
@@ -271,7 +271,7 @@ class ImageDataDeserializer : public DataDeserializerBase
         errno = 0;
         description.m_classId = strtoull(label.c_str(), &eptr, 10);
         if (label.c_str() == eptr || errno == ERANGE)
-            RuntimeError("Cannot parse label value on line %" PRIu64 ", second column, in file %s.", lineIndex, mapPath.c_str());
+            RuntimeError("Invalid map file format, must contain 2 or 3 tab-delimited columns, line %" PRIu64 " in file %s.", lineIndex, mapPath.c_str());
```
pkranen: Why '2 or 3', does it not allow n-dimensional regression (e.g. 4-dimensional for bounding box regression)? The square loss should be able to handle this.

Source/Readers/ImageReader/ImageDataDeserializer.cpp
DanielMerget: Good catch. Seems like a left-over of my debugging code. They can both be removed.

Tests/UnitTests/ReaderTests/Data/ImageReaderSimpleRegression_map.txt
```diff
@@ -0,0 +1,4 @@
+images\11.png	1	1
```
pkranen: Please add an end-to-end test that uses NaN valued labels for training and for testing. See existing E2E tests as examples.

Source/ComputationNetworkLib/TrainingNodes.h
```diff
@@ -44,11 +44,42 @@ class SquareErrorNode : public ComputationNodeNonLooping /*ComputationNode*/<Ele
...
         Value().VerifySize(1, 1);
-        Value().SetValue(v * v);  // Value = sum{ (I0[i] - I1[i])^2 }
+
+        // We scale the loss so that it is still comparable.
+        // E.g., if only half of the data was actually labeled, the loss will be doubled
+        // If we didn't do that, we would sometimes get very high losses, sometimes very low,
+        // without any chance to see why.  With scaling, the loss is always (roughly) at the
+        // same level, so the output is meaningful again. Thus, a lower loss means that the network
+        // is "actually better" and not just because the minibatch randomly contained very few labels.
+        const ElemType scale = max_labeled / num_labeled;
```
pkranen: What happens if num_labeled is zero?

Source/Readers/ImageReader/ImageDataDeserializer.h
DanielMerget: The lines you cited refer to the classification case. Regression can indeed be any dimension! The reason I had to change the message in this line was because the unit tests expected to see that exact wording in the message. The other variant is just as fine in my mind, but then the unit test would have to be adapted.

Source/ComputationNetworkLib/TrainingNodes.h
DanielMerget: The loss will become QNAN. I could add a conditional error-checking here, but on the other hand, does it really make sense for a label to be all nans?

Tests/UnitTests/ReaderTests/Data/ImageReaderSimpleRegression_map.txt
DanielMerget: I can do that. I have to admit though, that I messed up something in my python environment and cannot run the E2E tests anymore on my machine. I will have to fix that first, so it may take some time O:-)

Source/Readers/ImageReader/ImageDataDeserializer.h
pkranen: Fair enough, didn't check the full file.

Source/ComputationNetworkLib/TrainingNodes.h
pkranen: I agree that it is not meaningful to have all labels NaN, but currently nothing prevents a user from doing this. Your example was merging of two different databases or image sets. If something goes wrong or one adds new unlabeled images any label might be NaN by accident. The point is that the user should get an interpretable error message.

DanielMerget: You have a point there :-)

DanielMerget: I was just about to fix this in the deserializer, but then I realized that there may be scenarios where different readers are combined, e.g., ImageReader for features (with all-nan labels) and a TextReader for the actual labels, overwriting the old ones. So the check has to be postponed to the SquareErrorNode. Sadly, this will yield a less helpful error message since the location in the file and line will be lost. I just pushed the error-check to this pull request.

If you find that this should still be checked in the deserializer (for example, if labels that will be overwritten should be valid nevertheless, such as 0.0), this would be my implementation (already tested, insert before line 339 in ImageDataDeserialzer.h):

```
auto is_lowest = [](const PrecisionType& x)
{
    return x == std::numeric_limits<PrecisionType>::lowest();
};
if (std::all_of(result.cbegin(), result.cend(), is_lowest))
    RuntimeError("Encountered all-NAN label on line %" PRIu64 " in file %s.", lineIndex, mapPath.c_str());
```

Tests/UnitTests/ReaderTests/Data/ImageReaderSimpleRegression_map.txt
DanielMerget: Just added an E2E test. I hope it is illustrative (and sufficient). :-)

DanielMerget: Just saw that I should add one for training AND testing. The current test is doing it for training and writing, only. I'll add a test run, no problem. May take another hour or so.

DanielMerget: done

Source/ComputationNetworkLib/TrainingNodes.h
pkranen: The message in the SquareErrorNode is fine.

wolfma61:
Hi Daniel - 
we are still reviewing these changes.
We are also developing a sample to reach the same functionality through configuration and combination of CntkTextFormatDeserializer and ImageDeserializer, which is the original design to achieve this functionality. We try to minimize the amount of code to maintain, while matching the desired functionality. Please stay tuned ...

thx
Wolfgang

DanielMerget:
Hi Wolfgang,

Thanks for the info. Just to briefly give you my motivation for implementing regression in the ImageDeserializer directly (instead of simply using composition as you mentioned): One of my master's students (and I) are using ImageTransformers. However, in our scenario, i.e., landmark detection (and probably other scenarios as well), the transformations do not only affect the features, but also the labels,. In other words, labels have to be cropped and scaled in the same way as the image, etc. From our findings, this is difficult to implement if the labels do not originate from the same reader. We could be mistaken, though.

We are still experimenting with out implementation, but sooner or later I anticipate that ImageTransformers will work reliably on the label stream (if configured as such), and we would try to make it ready for a pull request as well.

Not sure whether that is something you want to support in the future, but at least it should be something for you to consider, so I thought I should mention it here.

Best,
Daniel

eldakms:
HI Daniel,

You should be able to combine Transformers on different deserializers independently. Please have a look at https://github.com/Microsoft/CNTK/wiki/Deserializers-and-Transforms

If you have i.e. feature images and label images then transforms can be specified independently on each of the streams.
Thanks!

DanielMerget:
Thanks @eldakms . I did not know that it was possible already. Is it possible to have the **very same** transformations for the streams (for example, with the **same seed**)? If so, how would I do that?

DanielMerget:
Remark @eldakms: Also, how can I apply a "crop transformation" to a vector of coordinates (not an image!)

eldakms:
Currently we only have transforms that work with images. You can implement your transform that crops any vector/matrix/tensor. Then it can be applied for any deserializer/stream. That would be a nice contribution!

Currently our stateful image transforms use random number generator and this can lead to different results. i.e. if you crop, cropping rectangles of the feature image1 and the corresponding label image1 can be different. Do you need absolutely the same transforms on all streams? This can be done, but we will need to slightly adapt the transform interface.
Thanks!

DanielMerget:
> Do you need absolutely the same transforms on all streams?

Yes, it is necessary for me because the streams are linked. In one of my scenarios, for example, I want to randomly crop face images (features) and adapt the landmark positions (labels) accordingly. In another setting, I try some "super-resolution" approach where the network learns how to increase the resolution of an input image. There, again, if I crop an input image (feature), the output image (label) has to be cropped in just the same way in order to still be correct. Of course, all of this also applies to other transforms as well.

The second setting is probably not too challenging to implement with different streams (at least both streams are images!). But for the first setting, I considered that the easiest - and yet probably clearest - way to achieve it was to adapt the ImageTransformers in such a way that they also (optionally) transform the label stream accordingly, if it was provided via the same mapping file (and therefore the same deserializer).

I hope that what I am trying to achieve makes more sense to you, now.

eldakms:
Understood. For now in case of two image streams you can specify "seed = bla" in a particular transform - then the applied transform should be absolutely the same. But you have to disable multithreading "numCpuThreads=1"(that will hurt performance) because currently transforms are shared between threads  .

For transforms that change vectors it could be done the same I suppose. I would prefer a model where we add new functionality and can compose it with the old one enabling new scenarios.

DanielMerget:
You don't happen to be located in Munich, Germany, by chance? I am working at TUM and noticed that part of the Microsoft CNTK team is from here, e.g., Wolfgang and Philipp. If you see any benefits on your part, maybe we could meet in person and discuss the details of how such a solution might look like. I will have to find a solution for my research anyway, so I would be happy to do it in such a way that it fits your expectations and/or quality standards to start from.

Given that this may be off-topic, feel free to contact me directly: https://www.mmk.ei.tum.de/~mer/

eldakms:
Yep, i'm also located in Munich.

DanielMerget:
Should I prepare another pull request that leaves out the changes in ImageDataDeserializer and only addresses the NAN-labels? Ignoring Unit/E2E-Tests, this is roughly ~50 lines of code. Would this be simple but valuable enough for you to be worth the trouble of future maintenance?

Best,
Daniel

eldakms:
Yep, would this work for CNTKTextFormat? I think Philipp wanted this change. Thanks!

DanielMerget:
Yes, I have already tested it on my machine with CNTKTextFormatReader, so it definitely works. I still have to adapt the Unit/E2E-test(s) accordingly. I will happily create a new pull request once I get to do that.

eldakms:
Great! Thank you so much!

DanielMerget:
Just submitted a PR for NaN-labels, see #823 :-)

Best,
Daniel

eldakms:
Great, thanks!

kassinvin:
@eldakms what do you mean "in case of two image streams you can specify "seed = bla" in a particular transform "? I have meet the same problem, but i don't understand. Look forward to your reply!

Issue #655, opened 2016-07-07
Convolution on Nvidia GTX1080 not converging


PeterGriffioen:
I have a problem with cntk GPU maintaining high errors throughout the run on the MNIST  02_Convolution.cntk example. I suspect this is related to https://github.com/Microsoft/CNTK/issues/644.

CNTK (gpu) on my workstation quite happily makes the 01_OneHidden.cntk example and other non-convolution models I've made. CNTK (cpu) works well with all models. 

I have followed all of the steps on the Setup for Windows page to the best of my ability, without installing the optional sections.
 
My workstation has a new Nvidia GTX1080 installed. It says it has cuda 8.0 installed no matter how many times I try to overwrite cuda 7.5 as per the instructions on the setup page. Regardless the cuda 7.5 program deviceQuery, which was compiled on the workstation, runs with a positive result.

CUDA Device Query (Runtime API) version (CUDART static linking)

Detected 1 CUDA Capable device(s)

Device 0: "GeForce GTX 1080"
  CUDA Driver Version / Runtime Version          8.0 / 7.5
  CUDA Capability Major/Minor version number:    6.1
  Total amount of global memory:                 8192 MBytes (8589934592 bytes)
MapSMtoCores for SM 6.1 is undefined.  Default to use 128 Cores/SM
MapSMtoCores for SM 6.1 is undefined.  Default to use 128 Cores/SM
  (20) Multiprocessors, (128) CUDA Cores/MP:     2560 CUDA Cores
  GPU Max Clock rate:                            1734 MHz (1.73 GHz)
  Memory Clock rate:                             5005 Mhz
  Memory Bus Width:                              256-bit
  L2 Cache Size:                                 2097152 bytes
  Maximum Texture Dimension Size (x,y,z)         1D=(131072), 2D=(131072, 65536), 3D=(16384, 16384, 16384)
  Maximum Layered 1D Texture Size, (num) layers  1D=(32768), 2048 layers
  Maximum Layered 2D Texture Size, (num) layers  2D=(32768, 32768), 2048 layers
  Total amount of constant memory:               65536 bytes
  Total amount of shared memory per block:       49152 bytes
  Total number of registers available per block: 65536
  Warp size:                                     32
  Maximum number of threads per multiprocessor:  2048
  Maximum number of threads per block:           1024
  Max dimension size of a thread block (x,y,z): (1024, 1024, 64)
  Max dimension size of a grid size    (x,y,z): (2147483647, 65535, 65535)
[...]

PR #771, opened 2016-08-19
Avoid double definition of atomicAdd on modern GPUs

DanielMerget:
NVIDIA finally implemented the double version of atomicAdd for Pascal. Hence the "custom" atomicAdd version must be excluded from compilation for these GPUs, otherwise it will be defined twice (and the compiler will complain).

Side note: @mahilleb-msft 's branch mahilleb/CuDnn5Test (recommended by @chrisbasoglu ) plus this small change made me able to compile and run CNTK successfully with a GTX1080 (Cuda 8.0 RC and cudnn 5.0).

See Issue #655.


n17s:
Daniel, this has been merged in the current master. 
Thank you!

PR #778, opened 2016-08-22
Linux build fix

junjieqian:
This patch fixes two issues encountered when building CNTK on Linux. The Linux version is "Ubuntu  4.4.0-28-generic", and the versions of the packages are same as advised.


zhouwangzw:
@junjieqian Thanks for your pull request. Which Ubuntu distribution version, and which gcc version are you using? 
We will look into the PR and come back to you.

junjieqian:
@zhouwangzw Thank you for quick response! The Ubuntu version is 16.04, and gcc is 5.1.3.

PR #822, opened 2016-09-07
G2P example fixed when running with CNTKTextFormatReader

hintz:
The G2P example wasn't running out of the box, and was lacking input data in ctf format. 
Details: _EmbedInput_, _EmbedLabels_ need a _Pass_ node as BeamSearchSequenceDecoderFrom expects a _labelSequence_ node during model editing.


cha-zhang:
I believe most of the PR contents has been fixed in the latest version. Closing this now.

PR #823, opened 2016-09-07
CNTKTextFormatReader and SquareError node can now handle NaN labels 

DanielMerget:
This PR comprises the NaN-label subset of PR #719 (slightly adapted for CNTKTextFormatReader) and includes a UnitTest for image regression via reader composition as well as an E2E-Test for image regression with NaN-Labels (again using reader composition).

Notes:
1) If the label vector has NaN-labels, it may not be modified (e.g., scaled) by .ndl or brainscript. This restriction would be gone _if and only if_ the labels would be represented by (Q)NANs internally, see PR #719 for details.
2) For the E2E-Test, I used the "old" baseline from PR #719 (i.e., the plaintext config shown in the log file would not work, but the results are still the same). If that is undesired, a new baseline should be created via --create-baseline/--update-baseline
3) The tests are currently written in .ndl O:-) (sorry)
4) Tested on Win8.1 and Win10 with Cuda8.0 RC and cudnn5.0 (it seems that GTX1080 is incompatible with Cuda7.5, so I cannot easily test it with Cuda7.5 myself). Since all tests pass, it should be fine, though.

Best,
Daniel


raaaar:
Hi Daniel, quick question -- could you elaborate on the issues with NaN handling in cuda? Have you experimented with CUDART_NAN(s)?

DanielMerget:
Hi,

First, I have to admit that I don't know too many details about CUDA, so the actual problem may be myself, after all.

I tried to use std::nan<> (in the reader/deserializer). Not sure to what exactly this is translated in CUDA. When I debugged in TrainingNodes.h, I could indeed verify that there were QNANs in the matrices, so that should (!) have been fine. However, I was somehow not able to "get rid of" those QNANs again. Since there is no "nan-check" member function for Matrix<>, I tried many different alternative approaches. The probably most direct or "promising" was to directly operate on the Matrix<> with things like x==x which should have given false for (Q)NAN and 1 otherwise (in my mind). However, that did not work out.

If something similar to the above would be possible, the code would not only work more robustly, but it would also be much cleaner, since the bulk of the work could be done in basically 1 line in TrainingNodes.h.

Back to your question: I never heard of CUDART_NAN(s). Are they different from the QNAN(s) I can debug with VS? If so, I probably should have used those CUDART_NAN(s), I suppose ...

raaaar:
Wrt to CUDART_NAN, take a look a this: http://stackoverflow.com/questions/15514286/way-to-get-floating-point-special-values-in-cuda

You can detect a NaN value using std::isnan(), same way it's done in Matrix::HasNaN(). However, this needs to be executed for each element, which tends to be rather slow. 

In general, your parsing code does make sense, but I'm not entirely convinced that returning lowest() value works for all use cases. Since the CTF reader is meant to be general purpose, this might have some unintended consequences when reading feature data. 

Ideally, if a sequence contains NaN values, it should be flagged by the reader as invalid, so that it can later on be ignored (replaced with a gap) by the packer when generating a minibatch layout. These missing sequences are then masked and excluded from the computation (take a look at Sequences::MaskMissingColumnsTo). The complete end-to-end implementation for this might take a fair amount of time.

DanielMerget:
> this might have some unintended consequences when reading feature data. 

I see your point that there is a danger for new files that try to (ab)use NaN. However, the reader would crash for NaNs right now, so there is no backward-compatibility issue.

How about some optional config parameter, e.g., "allowNAN = true" for the CNTKTextFormatReader/Deserializer, just to make 100% sure there will not be any side-effects unless explicitly asked for? I know this would not solve the principle issue, but is it maybe a tolerable solution?

I noticed the masking before, but I did not quite understand how/when/where the sequences are actually flagged earlier on. Could you point me to a place where this is currently done, so that I can get an impression of how it could be implemented in the TextReader?

raaaar:
"allowNaN" could be an acceptable stopgap if the reader actually produced NaNs.

A good place to start is SequencePacker, which actually generates layouts for all input streams. One quick thing you can try is to add a isValid flag to SequenceDataBase, check the flag in the SequencePacker::ReadMinibatch and if it's false, remove the corresponding sequences from the batch across all streams.

DanielMerget:
I just checked back in here. The reason there was no further progress so far is that the solution proposed by @raaaar is not suitable for the task. This is because Sequences::MaskMissingColumnsToZero(), as the name suggests, masks entire columns to zero rather than individual elements(*). In order to mask individual elements, there has to be a mask matrix (rather than a vector). My current approach uses a dense mask which is created from the nan-locations (std::min<>, see above) in the original input data.

If there are no NANs at all, any potential negative performance impact of NAN-handling could be easily avoided by an additional parameter, e.g., "allowNAN = true" (default: false), as suggested above. So performance should not be an issue.

Unless there is a neat way to convert NANs to a valid non-nan value _**on the GPU**_(**) (e.g., zero/one), I suppose the only alternative to my solution is to have the conversion done on the CPU (which is potentially slow). The upside would be that the implementation is "clean" and does not use std::min<> as a placeholder for NAN.

---------------------------
(*) To better understand the problem, this is how a label file (CNTKTextFormatReader) might look like:
...
|label 0.1 0.3 -0.2 NAN
|label -0.1 0.2 NAN 0.0
|label NAN -0.1 0.2 NAN
|label 0.3 NAN NAN NAN
|label 0.3 0.2 0.1 0.0
....

(**) one might expect that NAN == NAN yields 0, but instead it yields NAN in CUDA

raaaar:
Most of these changes have been included into 28dbe0b, but it looks like there's not a lot of demand for having NaNs in CTF format. Will keep the changes in a branch until somebody requests an integration. 

Thanks a lot for the contribution!

PR #824, opened 2016-09-07
Remove huge matrix dump

mfuntowicz:
Loading a model trained with an version before 1.7 like ResNet need to compute variance from invStdDev to match CuDNN v5.

The current commit remove the print of the invStdDev matrix to the console (which is really huge).


junjieqian:
Put into "#ifdef _DEBUG runInvStdDev.Print(); #endif" maybe better, as it provides an option for users in case they want it.

mahilleb-msft:
Thanks - Zhou also addressed this last week. It's was in fact a debugging left-over..

PR #849, opened 2016-09-18
fix CPPEvalClient build issue

weixsong:
This CPPEvalClient project could not build now, this maybe because the code structure reorganize.
Issues with build error:
* Could not find "Eval.h"
* Could not find EvalDll.lib

### DIFF
--- a/Examples/Evaluation/CPPEvalClient/CPPEvalClient.vcxproj
+++ b/Examples/Evaluation/CPPEvalClient/CPPEvalClient.vcxproj
@@ -30,6 +30,7 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <LinkIncremental>false</LinkIncremental>
     <OutDir>$(SolutionDir)..\..\$(Platform)\$(ProjectName).$(Configuration)\</OutDir>
+    <IncludePath>$(SolutionDir)..\..\Source\Common\Include;$(IncludePath)</IncludePath>
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
     <ClCompile>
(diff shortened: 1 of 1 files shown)


weixsong:
@zhouwangzw , Hi Zhou, I found that now this example could not build, will fix this and then rebase my example about extended interface example.

zhouwangzw:
Thanks @weixsong. Please let me know if you need any help. Pleaes note that in the latest master, the Eval examples in cntk.sln have been moved to the solution folder EndToEndTests\EvalClientTests\.

weixsong:
Hi @zhouwangzw , then <code>EvalClients.sln</code> in <code>CNTK\Examples\Evaluation</code> is a deprecated solution file?
Sorry for the misunderstanding.

weixsong:
Hi @zhouwangzw , I see that Eval examples have been move to EndToEndTest solution folder, and the project in End2EndTest still using <code>CPPEvalClient.cpp</code> cpp file, I think the solution in CNTK\Examples\Evaluation will cause some misunderstanding for user not familiar with CNTK architecture.

Make this project in CNTK\Examples\Evaluation\EvalClients.sln buildable will let it less confuse, will it bring some issue or unclear if we make this project buildable? or it will be removed in future?

zhouwangzw:
@weixsong Both the EvalClient.sln and CPP/CSEvalClient examples are shipped as a part of binary download package, so users who only downloaded binary package could use this solution file to directly build CPP/CSEvalClient examples, without creating their own solution and project files. This solution file has confiugrations customized for the layout of binary download package, so it is buildable with the binary download package, but not in the dev environment, where cntk.sln should be used. The cntk.sln contains project files in EndToEndTest/EvalClientTests to build these exapmples (same code but with different settings). The README.md files in both directories explain the purpose and usage of these projects. 

Thanks,
Zhou

PR #874, opened 2016-09-26
NCCL support in SimpleDistGradAggregator

nluehr:
Use NCCL collectives library (http://github.com/NVIDIA/nccl) to perform all reduce of 32- and 64-bit gradients when all MPI ranks 1) use GPUs and 2) reside on a common host.

Also use SimpleDistGradAggregator for 32- and 64-bit reductions even if CNTK was built with 1-bit SGD support.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,6 +147,14 @@ ifdef CUDA_PATH
     LIBS_LIST += cudnn
     COMMON_FLAGS +=-DUSE_CUDNN
   endif
+
+# Set up NCCL if needed
+  ifdef NCCL_PATH
+    INCLUDEPATH += $(NCCL_PATH)/include
+    LIBPATH += $(NCCL_PATH)/lib
+    LIBS_LIST += nccl
+    COMMON_FLAGS += -DUSE_NCCL
+  endif
 else
   DEVICE = cpu
 
@@ -313,6 +321,7 @@ MATH_SRC =\
 	$(SOURCEDIR)/Math/DataTransferer.cpp \
 	$(SOURCEDIR)/Math/RNGHandle.cpp \
 	$(SOURCEDIR)/Math/TensorView.cpp \
+	$(SOURCEDIR)/Math/NcclComm.cpp \
 
 ifdef SUPPORT_AVX2
 MATH_SRC +=\
--- a/Source/Common/Include/MPIWrapper.h
+++ b/Source/Common/Include/MPIWrapper.h
@@ -1,5 +1,6 @@
 //
 // Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
 // Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
 //
 
@@ -76,6 +77,7 @@ class MPIWrapper : public std::enable_shared_from_this<MPIWrapper>
     std::wstring m_myName;
     int m_numMPINodes;
     size_t m_numNodesInUse;
+    bool m_multiHost;
 
     // MPI communicator that reflects the current subset selection
     MPI_Comm m_currentComm;
@@ -149,6 +151,7 @@ public:
         MPI_Comm_rank(MPI_COMM_WORLD, &m_myRank);
         MPI_Comm_size(MPI_COMM_WORLD, &m_numMPINodes);
         m_numNodesInUse = m_numMPINodes;
+        m_multiHost = true;
 
         // Verify that the environment variable used by GetTotalNumberOfMPINodes()  
         // matches what the MPI API says. There're actually two possible cases:
@@ -309,6 +312,35 @@ private:
             fflush(stderr);
         }
         Ping("requestnodes (after change)");
+
+        // If all ranks run on a single host, we can enable optimized communication
+        // paths (e.g. NCCL). To determine if a single machine is being used, we
+        // check that MPI_Get_processor_name matches for all ranks.
+        const int nameMax = MPI_MAX_PROCESSOR_NAME + 1;
+        char myName[nameMax] = {0};
+        int  myNameLen = 0;
+        MPI_Get_processor_name(myName, &myNameLen) || MpiFail("requestnodes: MPI_Get_processor_name");
+        myName[myNameLen] = '\0';
+
+        std::vector<char> nameBuffer(m_numNodesInUse * nameMax);
+        char* allNames = nameBuffer.data();
+        MPI_Allgather(myName, nameMax, MPI_CHAR, allNames, nameMax, MPI_CHAR, m_currentComm)
+            || MpiFail("requestnodes: MPI_Allgather");
+
+        m_multiHost = false;
+        for(size_t i=1; i<m_numNodesInUse; i++)
+        {
+            if (strcmp(allNames, allNames+i*nameMax) != 0)
+            {
+                m_multiHost = true;
+                break;
+            }
+        }
+
+        fprintf(stderr, "requestnodes [%s]: using %d out of %d MPI nodes on %s (%d requested); we (%d) are %s\n",
+                msg, (int) m_numNodesInUse, (int) m_numMPINodes, m_multiHost ? "multiple hosts" : "a single host",
+                (int) requestednodes, (int) CurrentNodeRank(), IsIdle() ? "out (idle)" : "in (participating)");
+        fflush(stderr);
     }
 
 public:
@@ -362,6 +394,11 @@ public:
     size_t MainNodeRank() const
     {
         return 0;
+    }
+
+    bool IsMultiHost()
+    {
+        return m_multiHost;
     }
 
     // -----------------------------------------------------------------------
--- a/Source/Math/NcclComm.cpp
+++ b/Source/Math/NcclComm.cpp
@@ -1,0 +1,117 @@
+//
+// Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
+// Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
+//
+
+#include "NcclComm.h"
+
+#ifdef USE_NCCL
+#include "GPUMatrix.h"
+#include <nccl.h>
+#include <cuda_runtime.h>
+
+namespace Microsoft { namespace MSR { namespace CNTK {
+
+// allows to write cudaFunction() || "error"   (CUDA runtime)
+static void operator||(cudaError_t rc, const char *msg)
+{
+    if (rc != cudaSuccess)
+        RuntimeError("%s: %s (cuda error %d)", msg, cudaGetErrorString(rc), (int) rc);
+}
+
+NcclComm::NcclComm(int deviceId, const MPIWrapperPtr& mpi)
+    : m_ncclComm(nullptr), m_stream(nullptr)
+{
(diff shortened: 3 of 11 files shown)


Makefile
```diff
@@ -333,6 +341,7 @@ MATH_SRC +=\
 	$(SOURCEDIR)/Math/GPUWatcher.cu \
 	$(SOURCEDIR)/Math/GPURNGHandle.cu \
 	$(SOURCEDIR)/Math/MatrixQuantizerGPU.cu \
+        $(SOURCEDIR)/Math/NcclComm.cpp \
```
mahilleb-msft: Minor: could you use spaces here, as on previous line?
[Our Makefile has somewhat mixed space styles, which we'll fix; but let's try to be locally consistent.

Source/Common/Include/MPIWrapper.h
```diff
@@ -304,6 +308,45 @@ class MPIWrapper : public std::enable_shared_from_this<MPIWrapper>
...
+            fprintf(stderr, "requestnodes [%s]: using %d out of %d MPI nodes (single machine) (%d requested); we (%d) are %s\n",
+                    msg, (int) m_numNodesInUse, (int) m_numMPINodes, (int) requestednodes,
+                    (int) CurrentNodeRank(), IsIdle() ? "out (idle)" : "in (participating)");
+        }
+        else
+        {
+            fprintf(stderr, "requestnodes [%s]: using %d out of %d MPI nodes (%d requested); we (%d) are %s\n",
+                    msg, (int) m_numNodesInUse, (int) m_numMPINodes, (int) requestednodes,
+                    (int) CurrentNodeRank(), IsIdle() ? "out (idle)" : "in (participating)");
+        }
```
mahilleb-msft: Maybe you can roll that into one, using another ternary operator. (Like already existing for the last %s)

Source/Math/NcclComm.cpp
```diff
@@ -0,0 +1,93 @@
...
+        if (res != ncclSuccess)
+        {
+            RuntimeError("NcclComm failed to obtain ncclUniqueId");
+        }
+        MPI_Bcast(&ncclId, NCCL_UNIQUE_ID_BYTES, MPI_CHAR, 0, mpiComm)
+            || MpiFail("NcclComm: MPI_Bcase");
+
+        PrepareDevice(deviceId);
+        res = ncclCommInitRank(&m_ncclComm, numRanks, ncclId, mpi->CurrentNodeRank());
+        if (res != ncclSuccess) {
```
mahilleb-msft: Braces on separate line please (or omit for single-statement blocks)

Makefile
mahilleb-msft: (Actually: here, also use a hard tab, as on the previous line)

Source/Math/NcclComm.cpp
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
```
mahilleb-msft: Please also add this line below:
// Licensed under the MIT license. See LICENSE.md file in the project root for full license information.

(Regarding headers for existing files: please rebase one more time onto latest master, and add your additional line; we've added some headers where missing - thanks!)

Source/Math/NcclComm.h
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2016, NVIDIA CORPORATION. All reights reserved.
```
mahilleb-msft: Please also add this line below:
// Licensed under the MIT license. See LICENSE.md file in the project root for full license information.

Source/Math/NcclComm.cpp
mahilleb-msft: Please also add this line below:
// Licensed under the MIT license. See LICENSE.md file in the project root for full license information.

(Regarding headers for existing files: please rebase one more time onto latest master, and add your additional line; we've added some headers where missing - thanks!)

nluehr:
Changes incorporated and PR is re-based.

Source/Math/NcclComm.cpp
```diff
@@ -0,0 +1,95 @@
+//
+// Copyright (c) 2016, NVIDIA CORPORATION. All rights reserved.
+// Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
```
nluehr: Ah, I see. Yes, it get's compiled whenever CUDA is enabled. I meant to do something like this:

```
ifdef NCCL_PATH
        MATH_SRC += $(SOURCEDIR)/Math/NcclComm.cpp 
endif
```
But if you prefer that I move the include guards into NcclComm.cpp that would work too.

mahilleb-msft: Please include the guards in NcclComm.cpp, we follow that pattern elsewhere, too.

configure
```diff
@@ -963,6 +1005,7 @@ if test $enable_cuda = yes ; then
     echo GDK_NVML_LIB_PATH=$gdk_nvml_lib_path >> $config
     echo CUB_PATH=$cub_path >> $config
     echo CUDNN_PATH=$cudnn_path >> $config
+    echo NCCL_PATH=$nccl_path >> $config
```
mahilleb-msft: Could you make that
[ -z "$nccl_path" ] || echo NCCL_PATH=$nccl_path >> $config
so that we don't output NCCL_PATH if nccl_path is not defined?

nluehr: and done.

Source/Math/NcclComm.h
```diff
@@ -0,0 +1,56 @@
+//
+// Copyright (c) 2016, NVIDIA CORPORATION. All reights reserved.
```
mahilleb-msft: >reights [](start = 47, length = 7)

Typo: reights -> rights

Makefile
```diff
@@ -333,6 +341,7 @@ MATH_SRC +=\
 	$(SOURCEDIR)/Math/GPUWatcher.cu \
 	$(SOURCEDIR)/Math/GPURNGHandle.cu \
 	$(SOURCEDIR)/Math/MatrixQuantizerGPU.cu \
+	$(SOURCEDIR)/Math/NcclComm.cpp \
```
mahilleb-msft: To fix Windows build, please add the following into Source/Common/Common.vcxproj, line 65:
```
    <ClCompile Include="..\Math\NcclComm.cpp" />
```
(We'll need to figure out a better place in the end; but for now it seems the easiest change to fix Windows build)

mahilleb-msft: Also, please move this line up to to end of of two blocks up (insert after line 323, before ifdef SUPPORT_AVX2).
Otherwise, I'm seeing Linux build breaks for the non-CUDA compile.

Source/Common/Include/MPIWrapper.h
```diff
@@ -309,6 +312,36 @@ class MPIWrapper : public std::enable_shared_from_this<MPIWrapper>
...
+                m_multiHost = true;
+                break;
+            }
+        }
+
+        fprintf(stderr, "requestnodes [%s]: using %d out of %d MPI nodes on %s (%d requested); we (%d) are %s\n",
+                msg, (int) m_numNodesInUse, (int) m_numMPINodes, m_multiHost ? "multiple hosts" : "a single host",
+                (int) requestednodes, (int) CurrentNodeRank(), IsIdle() ? "out (idle)" : "in (participating)");
+        fflush(stderr);
+        Ping("requestnodes (after change)");
```
eldakms: >Ping("requestnodes (after change)"); [](start = 8, length = 36)

Should ping be with a different message? we already did the ping before.

Source/Math/NcclComm.cpp
```diff
@@ -0,0 +1,123 @@
...
+//
```
eldakms: Usually math should not contain any gpu related code. Only MathCuda should have it. I'm not sure what is your agreement with Mark though...

Source/Math/NcclComm.cpp
```diff
@@ -0,0 +1,123 @@
...
+    if (mpi->IsMultiHost())
+        return;
+
+    size_t numRanks = mpi->NumNodesInUse();
+    MPI_Comm mpiComm = mpi->Communicator();
+    std::vector<int> allDevs(numRanks);
+    MPI_Allgather(&deviceId, 1, MPI_INT, allDevs.data(), 1, MPI_INT, mpiComm)
+        || MpiFail("NcclComm: MPI_Allgather");
+
+    bool allRanksUseGPU = true;
```
eldakms: > bool allRanksUseGPU = true; [](start = 3, length = 28)

no need

Source/Math/NcclComm.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+    if (mpi->IsMultiHost())
+        return;
+
+    size_t numRanks = mpi->NumNodesInUse();
+    MPI_Comm mpiComm = mpi->Communicator();
+    std::vector<int> allDevs(numRanks);
+    MPI_Allgather(&deviceId, 1, MPI_INT, allDevs.data(), 1, MPI_INT, mpiComm)
+        || MpiFail("NcclComm: MPI_Allgather");
+
```
eldakms: should we also check that no two or more workers allocated to the same GPU device?

eldakms:
:shipit:

Source/Math/NcclComm.cpp
mahilleb-msft: Eldar: let's maybe talk tomorrow about this. There's no separate dir for MathCUDA, so the dir would be appropriate. The file however also contains the stub (lines 99 and below). On the other, there's a dependency on MPI (via the header), which normally MathCUDA / Math doesn't have. Maybe we can take a couple of minutes to see how we could untangle this.

---
In reply to: [87001650](https://github.com/Microsoft/CNTK/pull/874#discussion_r87001650) [](ancestors = 87001650)

Source/Math/NcclComm.cpp
```diff
@@ -0,0 +1,123 @@
...
+NcclComm::NcclComm(int deviceId, const MPIWrapperPtr& mpi)
+    : m_ncclComm(nullptr), m_stream(nullptr)
+{
+    if (mpi->IsMultiHost())
+        return;
+
+    size_t numRanks = mpi->NumNodesInUse();
+    MPI_Comm mpiComm = mpi->Communicator();
+    std::vector<int> allDevs(numRanks);
+    MPI_Allgather(&deviceId, 1, MPI_INT, allDevs.data(), 1, MPI_INT, mpiComm)
```
nluehr: For MPI_Allgather, sendcount and recvcount are per-rank values.

Source/Math/NcclComm.cpp
nluehr: Adding check. Note: this will disable NCCL when MPS is used as well.

nluehr:
All comments addressed in latest PR update, except where NcclComm source files are best located.

Source/Math/NcclComm.cpp
mahilleb-msft: I'm will be adding a work item on clean up here.

---
In reply to: [87059775](https://github.com/Microsoft/CNTK/pull/874#discussion_r87059775) [](ancestors = 87059775,87001650)

mahilleb-msft:
:shipit:

PR #877, opened 2016-09-28
support max utterance length in HTK Reader

ruizizhao:
support setting max utterance length in HTK reader when do training. 
This feature is just ported from HTK Kaldi reader.


mahilleb-msft:
Thanks, this is now in via a different commit from you, aa64648458a6c137e9b3e88d031f60596418fb4c.
Closing here...

PR #878, opened 2016-09-29
Updating README.md

MikalaiDrabovich:
Updating paths to examples in new format.


cha-zhang:
Hi @MikalaiDrabovich, we are in the process of updating the image examples into a new folder structure. It's almost ready, except moving some unit tests. Please wait a bit to see how the updated version works for you. 

If you are curious about the changes, check the branch chazhang/image_examples (still on-going work).

MikalaiDrabovich:
Actually, from here 
https://github.com/Microsoft/CNTK/tree/chazhang/image_examples/Examples/Image/MNIST/Config
it looks like the new paths from chazhang/image_examples branch will be exactly the same as proposed in the PR.

PR #950, opened 2016-10-26
Fix link on Readme.md

diegocrzt:
Fix the link for Language understanding, helpful when open multiple tabs of related works and have no idea of were could be the correct link :laughing:


alexeyo26:
Thank you for spotting this. I am updating ReadMe now and will add this change. Thank you once again.

diegocrzt:
it could be added in the meanwhile but ok, keep the open work  :+1:

PR #952, opened 2016-10-26
Adding a fast method to invoke the EvalWrapper for images

ant0nsc:
I've moved all the code to prepare the CNTK feature vector over to the unmanaged side, reducing time from 100ms to <1ms.


zhouwangzw:
@ant0nsc I have put some comments in the CodeFlow you sent out. After addressing them, the pull request can be integrated.

PR #997, opened 2016-11-04
Updated Docker files to include Python bindings

### LINKED ISSUE #979: Python Bindings out of the Docker box
Hey,
So I spent some time and got the python bindings of the GPU docker file going,
I found i had to install PCRE and jiggle the swig paths a bit. 
I was wondering if it would be possible to make the python bindings more plug and play in the docker file?

fiskio:
This allows the building of docker images that contain the python bindings.

Also added a docker file for CNTK-GPU-1bit-Image

Fixes #979

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1087,7 +1087,7 @@ python: $(ALL_LIBS)
             declare -A py_paths; \
             py_paths[34]=$(PYTHON34_PATH); \
             py_paths[35]=$(PYTHON35_PATH); \
-            export LD_LIBRARY_PATH=$$LD_LIBRARY_PATH:$$(echo $(LIBPATH) $(KALDI_LIBPATH) | tr " " :); \
+            export LD_LIBRARY_PATH=$$LD_LIBRARY_PATH:$$(echo $(GDK_NVML_LIB_PATH) $(LIBPATH) $(KALDI_LIBPATH) | tr " " :); \
             ldd $(LIBDIR)/* | grep "not found" && false; \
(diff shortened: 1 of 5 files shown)


n17s:
Thanks! It's in code review now...

n17s:
Hmm, I tried to build the 1bit Dockerfile and it could not find libnvidia-ml.so. After adding 
ENV  LD_LIBRARY_PATH /usr/local/cuda/lib64/stubs:$LD_LIBRARY_PATH 
to the Dockerfile I got it to build but when running I got this message

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
WARNING:

You should always run with libnvidia-ml.so that is installed with your NVIDIA Display Driver. By default
it's installed in /usr/lib and /usr/lib64. libnvidia-ml.so in GDK package is a stub library that is attached
only for build purposes (e.g. machine that you build your application doesn't have to have Display
Driver installed).
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Any ideas what's missing?

fiskio:
A word of warning, there is a little circular dependency here because the docker file clones a CNTK repository that need to include a small change in Makefile to include libnvidia-ml.so, so can't yet clone origin/master.
The docker files already point at the Microsoft repository, if you want to test it though you have to swap 'Microsoft' with 'fiskio' so that it clones my fork.

n17s:
The way I dealt with this was to temporarily add cuda/lib64/stubs into LD_LIBRARY_PATH for compilation, then remove those and add the /usr/lib ones after compilation is done.

fiskio:
Also updated the python environment setup to install all packages from `Scripts/linux/conda-linux-cntk-py34-environment.yml`. 
Added a few there too so that tutorials run smoothly.

n17s:
This has now been merged. Thanks for your contribution!

PR #1065, opened 2016-11-18
Use num_samples to accumulate metric_numer and metric_denom

saygintopkaya:
num_samples can be less than current_minibatch


KeDengMS:
Hi saygintopkaya,
In this sample current_minibatch should be the same as data[label_var].num_samples in the test loop. If not, there's a bug somewhere else. Can you put an assert in the loop to check that and let us know if it's being hit?

PR #1145, opened 2016-12-07
Fixing Docker CPU image to working with Speech example

AleksandrYahnev:
It is not possible to run example Speech/AN4/Config/FeedForward.cntk because of error `Gradient quantization is unsupported in CNTK binaries build without quantized gradient aggregation support`

Enabling 1bit-SGD https://github.com/Microsoft/CNTK/wiki/Enabling-1bit-SGD


n17s:
This cannot be accepted as is. You can fashion it after the GPU case where there are separate Dockerfiles for  1bit or not.

PR #1182, opened 2016-12-20
Fix an example

infwinston:
Fix an unmatched bracket in an example.
(Examples/Image/GettingStarted/06_OneConvRegrMultiNode.cntk)


mahilleb-msft:
Hi @infwinston, this one is merged, thanks for fixing it!

PR #1227, opened 2017-01-02
Update docstring for eval in functions.py

### LINKED ISSUE #1214: A mistake in the cntk python API document
In python API for CNTK 2.0beta7
The function eval of Class Function
image
image
image
The document says that the return value of function eval is `True` if updates have been performed.
While, in the source code of function eval, the return value is definely something different. It is the result of the function forward.
image

bencherian:
Fixed documentation for return value for "eval" in functions.py. Intended to fix #1214


n17s:
Thanks! This has been merged in.

PR #1257, opened 2017-01-06
Update CNTK_104_Finance_Timeseries_Basic_with_Pandas_Numpy.ipynb

epslong:
Fixing test_data p_up and p_down bug

### DIFF
--- a/Tutorials/CNTK_104_Finance_Timeseries_Basic_with_Pandas_Numpy.ipynb
+++ b/Tutorials/CNTK_104_Finance_Timeseries_Basic_with_Pandas_Numpy.ipynb
@@ -764,8 +764,8 @@
    "source": [
     "out = C.softmax(z)\n",
     "predicted_label_prob = out.eval({input:test_features})\n",
-    "test_data[\"p_up\"] = pd.Series(predicted_label_prob[:,0,0], index = test_data.index)\n",
-    "test_data[\"p_down\"] = predicted_label_prob[:,0,1]\n",
+    "test_data[\"p_up\"] = pd.Series(predicted_label_prob[0,:,0], index = test_data.index)\n",
+    "test_data[\"p_down\"] = predicted_label_prob[0,:,1]\n",
     "test_data['long_entries'] = np.where((test_data.p_up > 0.55), 1, 0)\n",
     "test_data['short_entries'] = np.where((test_data.p_down > 0.55) , -1, 0)\n",
     "test_data['positions'] = test_data['long_entries'].fillna(0) + test_data['short_entries'].fillna(0)"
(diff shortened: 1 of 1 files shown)


sayanpa:
Appreciate your effort to contribute and would love to take it in. However, the current version in the master is running and passing tests. 

Can you please describe the bug. Also, which version of the notebook are you using. With your suggested changes, the notebook is failing on latest build from master codebase.

epslong:
Thanks for following up quickly.

Here's the error I get from notebook:
[image: Inline image 1]

predicted_label_prob.shape gives (1,2243,2). Therefore, when using the
predicted_label_prob[:,0,0] and predicted_label_prob[:,0,1] to generate
p_up and p_down, it couldn't pass. Plus I think what what p_up and p_down
are meant for, they should be predicted_label_prob[0,:,0] and
predicted_label_prob[0,:,1]

I tried in my local cntk python environment as well, and it wouldn't pass
again due to the dimensional issue.











*Traceback (most recent call last):  File "<stdin>", line 1, in <module>
File
"C:\local\Anaconda3-4.1.1-Windows-x86_64\envs\cntk-py34\lib\site-packages\pandas\core\frame.py",
line 2419, in __setitem__    self._set_item(key, value)  File
"C:\local\Anaconda3-4.1.1-Windows-x86_64\envs\cntk-py34\lib\site-packages\pandas\core\frame.py",
line 2485, in _set_item    value = self._sanitize_column(key, value)  File
"C:\local\Anaconda3-4.1.1-Windows-x86_64\envs\cntk-py34\lib\site-packages\pandas\core\frame.py",
line 2656, in _sanitize_column    value = _sanitize_index(value,
self.index, copy=False)  File
"C:\local\Anaconda3-4.1.1-Windows-x86_64\envs\cntk-py34\lib\site-packages\pandas\core\series.py",
line 2800, in _sanitize_index    raise ValueError('Length of values does
not match length of ' 'index')ValueError: Length of values does not match
length of index*

So I doubt this has much to do with notebook.

Thanks

Grace

sayanpa:
I suspect there is a binary version installed on your machine and the array shape change related bug fix that went into the toolkit ( and the corresponding notebook change ) is in conflict. I have run the tutorial against the latest build from master and there is no regression. 

You can either build the binaries from the sources or wait till later in the coming week when we release the updated binaries.

sayanpa:
Hi Grace,
If you can verify on the latest release if the issue you pointed is resolved that would be highly appreciated. Would love to get a closure on the issue.

Thanks,
Sayan

sayanpa:
Hi Grace, if you can quickly verify on your setup with the latest beta branch would be nice if we can close this request.
Thanks,
Sayan

sayanpa:
Closing the request due to no update from the user for a while.

PR #1258, opened 2017-01-07
Add deviceId=localRank mode, enabling process select GPU by its local rank

FDecaYed:
This commit add localRank option. Using this, mpi processes can select GPU base on their local rank.

To use:
`cntk configFile=... deviceId=localRank`

For example, if we launch 3 processed on 2 nodes
Process 0 on node 0 with local rank 0 will will try to pick GPU 0 on node 0
Process 1 and 2 with local rank 0 and 1 will try to pick GPU 0 and 1 on node 1

User can also manually select GPU in this mode by tweak GPU order on each node with environment variable in advance. Normally this is not needed, since faster and closer GPUs may have smaller and closer numbers.

This commit should cover most of use cases when user know the topology. If other cases needed, we are willing to work on those later. 

Note that due to MPI implementation, windows version is not covered in this commit.

### DIFF
--- a/Source/CNTK/CNTK.cpp
+++ b/Source/CNTK/CNTK.cpp
@@ -1,5 +1,6 @@
 //
 // Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) 2017, NVIDIA CORPORATION. All rights reserved.
 // Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
 //
 // CNTK.cpp : Defines the entry point for the console application.
@@ -773,7 +774,17 @@ int wmainOldCNTKConfig(int argc, wchar_t* argv[])
 
     // full config info
     PrintBuiltInfo();
-    PrintGpuInfo();
+    if(EqualCI(val, "localRank")) {
+      auto data = GetGpuData(paralleltrain ? mpi->CurrentLocalNodeRank() : 0);
+
+      LOGPRINTF(stderr, "-------------------------------------------------------------------\n");
+      LOGPRINTF(stderr, "GPU info(only print info of gpu being used in localRank mode):\n\n");
+      LOGPRINTF(stderr, "\t\tDevice[%d]: cores = %d; computeCapability = %d.%d; type = \"%s\"; memory = %lu MB\n",
+		data.deviceId, data.cudaCores, data.versionMajor, data.versionMinor, data.name.c_str(), (unsigned long)data.totalMemory);
+      LOGPRINTF(stderr, "-------------------------------------------------------------------\n");
+      LOGPRINTF(stderr, "\n");
+    }
+    else PrintGpuInfo();
 
 #ifdef _DEBUG
     if (traceLevel > 0)
--- a/Source/CNTK/CNTK.cpp
+++ b/Source/CNTK/CNTK.cpp
@@ -690,7 +690,7 @@ int wmainOldCNTKConfig(int argc, wchar_t* argv[])
 
 #ifndef CPUONLY
     ConfigValue val = config("deviceId", "auto");
-    if (!EqualCI(val, "cpu") && !EqualCI(val, "auto"))
+    if (!EqualCI(val, "cpu") && !EqualCI(val, "auto") && !EqualCI(val, "localRank"))
     {
         if (static_cast<int>(val) >= 0) // gpu (id >= 0)
         {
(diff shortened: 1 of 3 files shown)


FDecaYed:
Looking at make some further change to reduce GPU initialization time. Will update soon.

FDecaYed:
Hi,

I just added one commit that optimized the path to acquire gpu information.

When not using new localRank mode, everything should remain same behavior with checking support for gpu faster. That is one of the 2 places where getAllGpusData is called and causing slowdown on multi gpu case.
When using localRank mode, cntk will only print information of the gpu being used for each rank and check support for that gpu. Which means all need for getAllGpusData is removed.

Time to initialize 8 gpus gose from minutes to less than a second in localRank mode.

Deyu

FDecaYed:
Hi @KeDengMS 

Just for your reference, the crash you discussed with Nathan is very similar to what I changed in this PR in `GpuData GetGpuData(DEVICEID_TYPE deviceId)`

You can probably do something similar in `DeviceDescriptor::GPUDevice()` to validate gpu by ID and avoid calling `AllDevices()`(with `getAllGpusData()` inside that do the reset).
Or if this got merged, you can even call this new `GetGpuData()` in `DeviceDescriptor::GPUDevice(`) if you don't want to expose cuda api to CNTK/Source/CNTKv2LibraryDll/Common.cpp.

Whether to do something about the way cntk get information from GPU(`getAllGpusData()`) and select GPU (essentially the `cudaMemGetInfo()` needed that cause all these) may be a wider question here overall.

Source/Common/Include/MPIWrapper.h
```diff
@@ -376,6 +376,20 @@ class MPIWrapper : public std::enable_shared_from_this<MPIWrapper>
     {
         return m_myRank;
     }
+
+    size_t CurrentLocalNodeRank()
+    {
+        const char* p = std::getenv("OMPI_COMM_WORLD_LOCAL_RANK");
```
KeDengMS: Does this work on Windows with msmpi?

KeDengMS: Does OMPI_COMM_WORLD_LOCAL_RANK work on Windows with msmpi?

Source/CNTK/CNTK.cpp
```diff
@@ -690,7 +691,7 @@ int wmainOldCNTKConfig(int argc, wchar_t* argv[])
 
 #ifndef CPUONLY
     ConfigValue val = config("deviceId", "auto");
-    if (!EqualCI(val, "cpu") && !EqualCI(val, "auto"))
+    if (!EqualCI(val, "cpu") && !EqualCI(val, "auto") && !EqualCI(val, "localRank"))
```
KeDengMS: How does this work with multi-tenant cases? Say, we have a 8-GPU server running 2 4-GPU jobs. Each job would have local rank 0-3, so GPU 4-7 is not used?

FDecaYed: User will have two choice here.
1. use 'auto' mode in at least one of the 2 jobs.
2. if they know their machine well, they can run a job on any combinations of gpu they want manually with CUDA_VISIBLE_DEVICES and localRank mode

Source/Common/Include/MPIWrapper.h
FDecaYed: I stated this will not work with windows and asked amit about the windows alternative of this. Seems there is no such implementation as far as we know in msmpi.

KeDengMS:
@FDecaYed My understanding is that this PR fixes the slow start-up in BestGPU::Init when CNTK used to call cudaSetDevice. Now that we've removed it, do you think we still need localRank for GPU selection?

FDecaYed:
@KeDengMS Sorry for the delay, just got back from vocation.
It is still needed in some cases where we want to specify GPU selection manually. There is no way now to do that in auto mode, which will give us non-deterministic selection.

anwald:
On a typical high-end machine now, you get 2 sockets with a numa topology. The trajectory looks like we will soon go to 4 socket servers in not too long. Hence it is in our interest to pin the process to a GPU whose PCIE host bridge is on the same die as the cores to which the process is affinitized. For this reason, snapping by rank is fairly naïve in a multi-socket architecture. Also, the proposed solution assumes no contention with other consumers of the machine, which at least for us in incorrect when an upfront investment has been made in a multi-GPU machine. There is an example here https://www.open-mpi.org/faq/?category=runcuda of using hwloc to find local gpu's. Maybe we could then make the gpu assignment also a function of rank and the result of hwloc, so that we could deterministically prevent assignment collision. The remaining hurdle to overcome is to try to prevent process migration by the operating system, which maybe we can do by system calls from within the cntk process, otherwise using taskset outside the cntk process. Interested in getting more thoughts on these issues.

FDecaYed:
@anwald This PR is purely for enabling a way to manually select GPU with CUDA_VISIBLE_DEVICES in cntk. It is not attempt to solve entire GPU selection problem, and this is exactly the reason I made it a separate mode.

I agree with your idea that we should get more thoughts on gpu selection question. Or more sepcifically, improve the 'auto' mode in CNTK. To me, it is more of an open ended question. A single right answer is hard, but improve it over time base on use cases is feasible.

Here is some random thoughts I have in addition to yours:
Topology other than '2 socket with symmetric PCIE'?
Other type of link(e.g. nvlink) between GPUs?
GPUs in a sever have different compute/memory capability?
If assume contention, what is the policy?
Which is more important, CPU-CPU, CPU-GPU or GPU-GPU distance? Or even disk?
Multi-Node?
How does containerize affect this? Could some of the parts be done out side of cntk?

FDecaYed:
Further discussion around this problem needed, but problem addressed by this PR is already gone. Closing.

PR #1264, opened 2017-01-08
Build fix for Linux

ivannp:


junjieqian:
This is caused by different GCC version. And it is `int32_t` (`DeserializerType`).

ivannp:
ubuntu 16.04, g++ 5.4.

It's a warning, but I think the build is treating most warnings as errors.

mahilleb-msft:
Thanks @ivannp - going to take this PR, but it requires a bit of time to integrate.
Best, Mark

PR #1272, opened 2017-01-10
FastRCNN - Bad bonding box read when there is only one ground truth annotation

nadavbar:
This pull request contains a fix to the following issue:
When an image has only one ground truth annotation, the ground truth bounding box is wrongly read due to the fact that numpy loads the *.bboxes.tsv file as a 1D array (instead of a 2D). This leads to the bounding box of the GT region to be treated as [x1 x1 x1 x1], which leads to all of the ROIs being marked as background (since there are no overlaps with the ground truth).

@pkranen  - Can you please have a look? Let me know if any change to the code is required.


pkranen:
Thanks for fixing!

Issue #1268, opened 2017-01-09
CNTK GPU console output goes neither to stderr nor to stdout


jakrivan:
Simple repro on latest version:

`K:\CNTK-2-0-beta7-0-Windows-64bit-CPU-Only\cntk\cntk>CNTK.exe 1>out.txt 2>err.txt`

err.txt contains build info as expected

`K:\CNTK-2-0-beta7-0-Windows-64bit-GPU\cntk\cntk>CNTK.exe 1>out.txt 2>err.txt`

out.txt and err.txt are empty.

beta 6 exhibits same behavior. I do not know about older 2.0 versions; but 1.7.2 and older versions did not exhibit this issue (any SKU of cntk was properly logging to stderr).

Is this expected or/and is there any workaround how to capture the cntk output?

avader906:
You can specify the output redirection via cntk config file:
```
stderr = "$OutputDir$/yourlogfilename.log"
```
supplying the name and the path for the log file from the prompt would work as well:
```
K:\CNTK-2-0-beta7-0-Windows-64bit-CPU-Only\cntk\cntk>CNTK.exe configfile=yourconfig.cntk command=yourCommand stderr="yourpath\filename"
```

jakrivan:
Thanks for mentioning the log redirection.
I showed the issues on redirecting to files just to demonstrate the issue (regression and only specific to GPU SKU).
We are redirecting programatically (opening stderr and stdout streams and reading from them) - so redirecting to files is not an option for us

jakrivan:
I might have found a source of the issue (not the actuall rootcause, but at least way how to fix it):

Stderr might not be allways flushed by the Cpp runtime shipped with VS 2013:

https://connect.microsoft.com/VisualStudio/feedback/details/1441507/stdout-and-stderr-not-always-flushed-upon-process-exit

CNTK makes good efforts in flushing stderr by itself (TerminateHandler; flushing on errors etc) - with one exception - exit after printing build info if no commandline args are supplied:

https://github.com/Microsoft/CNTK/blob/master/Source/CNTK/CNTK.cpp#L863

So I believe adding 'fflush(stderr);' should do the trick.

I still do not fully inderstand why this manifests only on GPU SKU of 2.0 beta builds (and any CPU nor any older 1.x builds) - it might as well be some timing coincidence where runtime manages to properly flush the stderr

PR #1274, opened 2017-01-10
Flushing stderr on no cmd args given

jakrivan:
Addresses issue described in https://github.com/Microsoft/CNTK/issues/1268
Stderr not properly flushed when no cmd args given (affects just GPU skus)


mahilleb-msft:
Thanks, I'm looking at merging this.

PR #1311, opened 2017-01-18
A couple of bugs fixed in CNTK_203_Reinforcement_Learning_Basics tutorial

eveselov:
a) The order of loss function arguments was incorrect, causing crash in training
b) env resetting moved from outside to inside the loop to allow visualization on all iterations
c) env rendering uncomented to allow visualization of the model


sayanpa:
I am looking into the code change and will merge with the base after running regression tests. Thanks for your contribution.

sayanpa:
Quick update, the build failed on a regression test. We will have to take a closer look into the reason for the failure.

PR #1378, opened 2017-01-30
Declare deleted functions public in C++ API header file.

taeguk:
Declare deleted functions not private but public for better error messages.

Scott Meyers is saying in 'Effective Modern C++' (Item 11, page 75/76): 
>By convention, delete function are declared public, not private. ... will result in better error messages. There's a reason for that. When client code tries to use a member function, C++ checks accessibility before deleted status.

When a user tries to use a private deleted function, compiler complain about the function being private.
So, declaring deleted functions as not private but public results in better error messages.


ebarsoumMS:
Can you rebase to latest master?

taeguk:
@ebarsoumMS I resolved merge conflicts. This PR can be merged now.

ebarsoumMS:
Sorry, it is by design to keep them private.

PR #1399, opened 2017-02-04
Added ability to change progress printer metric treatment

bencherian:
I added an additional (optional) keyword parameter to allow progress_printer to not treat the metric as a percentage. This will keep it from multiplying the metric by 100.0 in output in cases where the metric is not intended to be treated as a percentage.


bencherian:
Sorry...forgot to push to fork before submitting PR.

PR #1420, opened 2017-02-08
Pack gradients into continous buffer

junjieqian:
As discussed, this design is to pack the gradients (size less than thresholdsize) into a continuous buffer for better performance.
@nluehr, please review.


nluehr:
I believe the headers originate in CPU memory, so using NCCL to broadcast them is probably slower than just using MPI.

For the gradients, the aggregation scheme causes an extra cudaMemcpy to get the data from the computation node into the aggregated buffer. These copies will be latency bound just like the small ncclAllReduces were to begin with, so I don't expect a large performance improvement. It would be better to re-allocate the nodes underlying gradient storage.

junjieqian:
@nluehr, thanks for the comments! 
I will remove the NCCL bcast for headers.
Reallocate the nodes to a continuous buffer instead of packing/copying them would be the next step as discussed.

PR #1437, opened 2017-02-10
nluehr/examplefixes

nluehr:


chrisbasoglu:
This change can't be merged as is.  It causes test: char_rnn_test.py from executing in 27 seconds, to running longer (or hanging) than 3 hours:

do the following with and without your change:

cd Tests/EndToEndTests/CNTKv2Python/Examples
pytest char_rnn_test.py

nluehr:
Sorry about that. I didn't see the test.

PR #1441, opened 2017-02-11
Fix a dangling pointer problem in CNTKv2LibraryDll.

taeguk:
There is a dangling pointer problem in code below.
```cpp
...
CNTK::Variable elementVariable;
CNTK::FunctionPtr functionPtr;
{
    CNTK::FunctionPtr funcPtr = CNTK::ElementDivide(leftOperand, rightOperand);

    elementVariable = funcPtr;
    // functionPtr = funcPtr;     // (A)

    elementVariable.Owner();   // not error.
}
elementVariable.Owner();       // error. (dangling pointer)
                               // but not error when (A) is not commented.
...
```

The reason is that ```m_ownerFunction``` is a raw pointer in ```VariableFields```.
https://github.com/Microsoft/CNTK/blob/fd2e7969a56777345fc8b7007b03e95edc35036f/Source/CNTKv2LibraryDll/Variable.h#L21

I know that it is for resolving circular reference problem.
But raw pointer is very dangerous. In this case, using std::weak_ptr is good practice.
The std::weak_ptr does not affect reference count like raw pointer. But the std::weak_ptr can detect whether pointer is valid or not.

I fix a dangling pointer problem using std::weak_ptr.

### DIFF
--- a/Source/CNTKv2LibraryDll/API/CNTKLibrary.h
+++ b/Source/CNTKv2LibraryDll/API/CNTKLibrary.h
@@ -1629,7 +1629,7 @@ private:
 
         CNTK_API static Variable Deserialize(const Dictionary& dictionary, const ::CNTK::DeviceDescriptor& device = DeviceDescriptor::UseDefaultDevice());
 
-        void SetOwner(Function* ownerFunction);
+        void SetOwner(const std::weak_ptr<Function>& ownerFunction);
 
         Variable CompositePreservingCopy(const std::shared_ptr<const Function>& composite) const;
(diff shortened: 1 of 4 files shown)


zhouwangzw:
@taeguk Thanks for your suggestion. Using weak_ptr sounds a good solution. I am rebasing your PR to the latest master and will merge after tests. 
Thanks for your effort to make CNTK better.

zhouwangzw:
@taeguk We have merged your change into master. Thanks again for your contribution!

PR #1444, opened 2017-02-12
Small language change to comments

jerogers-msft:
I think you meant "Set epoch size to a larger number for lower training error" rather than "Set epoch size to a larger number of lower training error"
Thanks for the tutorials!


chrisbasoglu:
Thank you

Issue #1178, opened 2016-12-16
Not working randomization


snurkabill:
Hi!

We recently noticed that CNTK does not randomize and for same CNTK config we get the (absolutely) identical error function value. This is not expected from SGD. I also tried dropout (which should be also randomized) and it resulted in the same results. 
So it seems to me that there is: 

no random weights initialization
no randomized input vectors
no randomized dropout 

Maybe some bug about different random seed?  (CNTK version 2.0.6 beta CPU only) 

I provide my CNTK config file with output log of CNTK for completenes.  

CNTK CONFIG:
===========================================================================================

RootDir = /root
deviceId = -1
numCPUThreads = 8
shareNodeValueMatrices = true
command = train
precision = "float"
modelPath = "$RootDir$/CNTKModel"
traceLevel = 1
numMBsToShowResult = 500
initOnCPUOnly = true

```
train = [
    action = "train"
	    BrainScriptNetworkBuilder = [
		# number of features
        featDim = 685
        # number of outputs
        labelDim = 3

features = Input (featDim)
labels = Input (labelDim)

_hiddenDim1 = 32
_hiddenDim2 = 16
_hiddenDim3 = 8

_denseLayer1 = DenseLayer {_hiddenDim1, activation=ReLU, initValueScale=1, init="uniform"} (features)
_denseLayer1drop = Dropout (_denseLayer1)
_denseLayer2 = DenseLayer {_hiddenDim2, activation=ReLU, initValueScale=1, init="uniform"} (_denseLayer1drop)
_denseLayer2drop = Dropout (_denseLayer2)
_denseLayer3 = DenseLayer {_hiddenDim3, activation=Sigmoid, initValueScale=1, init="uniform"} (_denseLayer2drop)
_denseLayer3drop = Dropout (_denseLayer3)
_mojeVrstvicka = LinearLayer {labelDim, initValueScale=1, init="uniform"} (_denseLayer3drop)
_outputLayerPotential = Sigmoid (_mojeVrstvicka)
_lossFunction = SquareError (labels, _outputLayerPotential)
_errors = ErrorPrediction (labels, _outputLayerPotential)

featureNodes    = (features)
labelNodes      = (labels)
criterionNodes  = (_lossFunction)
evaluationNodes = (_errors)
outputNodes     = (_outputLayerPotential)
    ]

SGD = [
epochSize = 0
maxEpochs = 2
minibatchSize = 1
dropoutRate = 0.5
[...]

PR #1448, opened 2017-02-13
Adding useRandomDropout option into SGD block

jakrivan:
This is to add an optional randomization of dropout process.
Should address issue https://github.com/Microsoft/CNTK/issues/1178

### DIFF
--- a/Source/ComputationNetworkLib/ComputationNetwork.cpp
+++ b/Source/ComputationNetworkLib/ComputationNetwork.cpp
@@ -573,13 +573,20 @@ template <class ElemType>
 
     list<ComputationNodeBasePtr> rngUserNodes = net->GetNodesWhere(nodeIsIRngUser, node);
 
+    if (randomDropout)
+    {
+        //this is not ideal - see http://www.cplusplus.com/faq/beginners/random-numbers/
+        // however still get's some form of randomness compared to fixed seed
+        srand((unsigned int)time(NULL));
+    }
+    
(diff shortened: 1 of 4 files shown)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

chrisbasoglu:
Thank you for your contribution.  

This change (when enabled) unfortunately adds non determinism by using a time based random seed and will break checkpointing as well as reproducibility.

In the CNTK v2 APIs, we allow users to explicitly set the random seed for the dropout nodes which allows achieving the desired results without losing reproducibility and determinism. If the user wants this functionality in V1, a better way would be add a new randomSeedBase option in the config that will be added to the seed base used for RNG nodes and give users explicit control over varying the randomness across networks and runs if desired, without losing reproducibility and determinism.

jakrivan:
Hi chrisbasoglu - thanks for your feedback.

The reproducibility and determinsm was the exact reason why 'useRandomDropout' was made optional and with default being false. That way you can remove this from your config and keep the deterministic behavior.

I agree that having the seed in config is even better though - I'll try to find some spare time to alter the code to this.

In any case - can you please point me to the code or documentation for the behavior you mentioned:

> In the CNTK v2 APIs, we allow users to explicitly set the random seed for the dropout nodes

This might be quite helpful to us.

Thanks
Jan

chrisbasoglu:
Sorry we are adding this feature in right now.  Look for it in master in a few days.

Chris

PR #1450, opened 2017-02-13
Add another check for end of epoch.

FDecaYed:
Currently, CNTK has strange behavior at the end of epoch when loading sequence data.

Assume reading lots of sequences of length 3. If we set `epochSize = 100`, `minibatchSize = 90`, we get
```
minibatch 1:90 samples
minibatch 2:9 samples
minibatch 3:3 samples
Samples in epoch: 102
```
Following is CNTK definition of `minibatchSize` and `epochSize`:
> Despite our clear definition of `minibatchSize` being the number of samples between model updates, there are two occasions where we must relax the definition:
> 
> sequential data: Variable-length sequences do not generally sum up to exactly the requested minibatch size. In this case, as many sequences as possible are packed into a minibatch without exceeding the requested minibatch size (with one exception: If the next one sequence in the randomized corpus exceeds the length of the minibatch size, the minibatch size will consist of this sequence).
> 
> data parallelism: Here, the minibatch size is approximate, as our chunk-based randomization algorithm cannot guarantee that each worker receives precisely the same number of samples.
> 
> All of the above considerations also apply to `epochSize`.

New behavior after this change:
```
minibatch 1:90 samples
minibatch 2:9 samples
Samples in epoch: 99
```

Note that there might be similar behavior need further investigation in 'end of sweep' check and in NoRandomizer.

### DIFF
--- a/Source/Readers/ReaderLib/BlockRandomizer.cpp
+++ b/Source/Readers/ReaderLib/BlockRandomizer.cpp
@@ -121,12 +121,12 @@ Sequences BlockRandomizer::GetNextSequences(size_t globalSampleCount, size_t loc
     {
         assert(globalSampleCount > numGlobalSamplesLoaded && localSampleCount > numLocalSamplesLoaded);
         bool atTheSweepBoundary = result.m_endOfSweep;
-        // in case when we continue filling up a minibatch that crosses a sweep boundary, 
+        // in case when we continue filling up a minibatch that crosses a sweep boundary,
         // make sure that it does not exceed the required number of samples. Set the atLeastOnceSequenceNeeded
         // flag to false.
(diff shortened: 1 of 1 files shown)


raaaar:
The reader behavior follows the spec -- only a minibatch consisting of a single sequence is allowed to exceed the minibatch/epoch size. This is not to say that the spec is correct, it seems we have a definitional problem. However, the proposed check in no way addresses the issue, instead it simply sets the 'end-of-epoch' flag depending on the requested (not actual) minibatch size.

raaaar:
Closing this PR, as the issue pointed out here has been address by 39464485a2fe037e1f764425536e9ff47e9917bb.

PR #1516, opened 2017-02-22
Fix for Stochastic CUDNN_STATUS_EXECUTION_FAILED

FDecaYed:
There are two main changes in this PR.
1. `cudnnFindConvolutionXXXAlgorithmEx` is now used instead of `cudnnFindConvolutionXXXAlgorithm`. This change will avoid the failed malloc inside Find, which is likely the cause of cudnn failure 8:`CUDNN_STATUS_EXECUTION_FAILED`
2. All cudnn convolution calls now uses a shared workspace. This will allow better memory usage control and help enabling faster algorithm.

There are 3 noticeable improvement testing with Resnet50/Titan XP. Note that the test script I use is not up to date, but similar behavior should be expected with current CNTK stock script.
1. TOT master can run at most with `minibatchsize=112`, and then crush with `CUDNN_STATUS_EXECUTION_FAILED`. This PR allow to run up to `minibatchsize=135` with full speed and then crush with `CUDA failure 2: out of memory`. This is the expected error since we are really reaching the 12GB memory limitation in this case.
2. During initialization phase, TOT uses a lot more memory than after initialization(running phase). While this PR ensure a steady memory foot print(running phase memory usage is the max usage).
3. After 100MB memory is reseved for GPU to run correctly, CNTK now will use `min(free memory, 1GB)` as a shared workspace for convolution. Overall, we lose some memory when small batchsize is used(up to 1GB, usually several hundred MB), but for larger batchsize, we save memory. For Resnet 50 from `minibatchsize=64`, we start to see memory saving(TOT:7124MB vs PR:7058MB).

Other things need discussion:
- Cleaning up old workspace API because they are no longer needed
- Cudnn **DOES NOT** assume that same workspace and algorithm will work when reducing batch size, it is not safe to assume that. We need to change that part
- Expose the 1GB memory limit to user via `maxTempMemSizeInSamples`

### DIFF
--- a/Source/Math/CuDnnConvolutionEngine.cu
+++ b/Source/Math/CuDnnConvolutionEngine.cu
@@ -1,5 +1,6 @@
 //
 // Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) 2017, NVIDIA CORPORATION. All rights reserved.
 // Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
 //
 
@@ -166,9 +167,9 @@ private:
 
 enum class AutotuningState : int
 {
-    Init = 0,          // initial state 
-    PendingTuning = 1, // memory of all nodes have been allocated, it's safe to do tuning now 
-    Running = 2        // done tuning, no long performing auto-tuning, code is running normally 
+    Init = 0,          // initial state
+    PendingTuning = 1, // memory of all nodes have been allocated, it's safe to do tuning now
+    Running = 2        // done tuning, no long performing auto-tuning, code is running normally
 };
 
 template <class ElemType>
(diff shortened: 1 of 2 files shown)


FDecaYed:
Update:
1. Remove all dependency on workspace that passed in and use static m_workspace instead. workspace is still passed since the API is shared by other Engines.
2. Changed code and addressed comment on algorithm selection behavior when minibatchsize decrease.
3. Loop over algorithms to get memory size needed to choose the best algorithm. This can potentially allow better algorithm to be selected compare to the 1GB approach. Down side is this require more memory at tuning stage(2nd minibatch). The memory need for tuning is very similar to Top Of Tree master now. While after tuning, the workspace is resized to what actually needed and the memory usage is even lower now.

amitaga:

amitaga:
@chazhang is added to the review. #Closed

amitaga:
@Philly is added to the review. #Closed

FDecaYed:
I just fixed a little bug when calculating workspace size when it reaching to boundary.

FDecaYed:
just fixed a bug where I forgot to change the old matrix initialization into resize when I copy intialization to somewhere else.

Issue #1456, opened 2017-02-14
Metric Output in Progress Printer


bencherian:
The metric output by the ProgressPrinter class is currently always treated as a percentage value. While this makes sense for classification problems, it does not make sense for regression problems in which the eval error may just be an alternative loss function (like absolute error instead of squared error). Can we add the option to not treat the metric as a percentage. I submitted PR #1520 to add an optional keyword argument that would allow for such a treatment, but I recently saw that there is a branch in the repository (vit-stepanovs/progress-api-2) which is being used to refactor the ProgressPrinter class into an abstract base class and separate concrete classes for TensorBoard and text based logging. However, they still are hard coded to treat the metric as a percentage in the output. Could a similar change be made to allow for treatment of the metric as an absolute value instead of a percentage?

PR #1520, opened 2017-02-24
Allow for non-percentage metric in ProgressPrinter

bencherian:
Added optional argument to constructor for ProgressPrinter which enables treatment of metrics as standard values.
Original treatment was hardcoded to output metric as a percentage. The updated class will output metric values in the same format as the loss if metric_is_pct is set to false. (The default value is True to preserve backwards compatibility.) Provides solution for #1456


bencherian:
Reopening to bump cla bot.

bencherian:
Any chance of getting this merged? If this is considered an unsuitable addition, that's ok...but if it's going to get merged, it would be nice if it could be in the next beta release.

PR #1582, opened 2017-03-08
Enable NCCL support for Python in V2 gradients aggregation

junjieqian:
@nluehr , can you please take a look at this commit?


nluehr:
Re-initializing NCCL at ever call to AggregateImpl() in Source/CNTKv2LibraryDll/DistributedCommunicator.cpp:308 is very expensive. It would be much better to make m_nccl a member of MPICommunicatorImpl, and set it up once for all.

junjieqian:
Hi Nathan @nluehr , thank you for your suggestion. I will to find a way to pass the deviceID to MPICommunicationImpl, because it is called without deviceID in most of cases.

PR #1644, opened 2017-03-22
fix workspace corner case and reduce memory footprint

FDecaYed:
This fixes several corner case that lead to error when batchSize changes frequently.

I also found a way to reduce the workspace size after everything is settled, which will reduce the running phase memory usage. This will avoid cntk error out with 'out of memory' when workspace is allocated and some other matrix are resized to larger size.

I also tested with master so this commit can be merged into master without issue.


KeDengMS:
Many thanks for the fast investigation and fix!

PR #1654, opened 2017-03-24
Fix corruption in multi-GPU training

nluehr:
Makes NcclComm's stream sync with default stream in order to properly respect device-to-device transfers from gradients to merged buffer.


junjieqian:
Hi @nluehr , does this change require another version of NCCL (other than nccl-1.3.0)?

nluehr:
No. It is independent of NCCL version.

amitaga:

PR #1714, opened 2017-04-07
[WIP] Refine CNTK build process.

weijianwen:
I am trying to refine the build process of CNTK. This commit enble use of pre-installed `multiverso` and `1bit-sgd` libraries outside CNTK source tree. More specifically, it

1. checks `multiverso` and `1bit-sgd` avaialbility during `configure` rather than `make`. 
2. adds `--with-multiverso=` and `--with-1bitsgd=` flags for specifying lib locations oustside CNTK source tree.
3. Refine `Makefile` to adapt path chages for multiverso and 1bisgd libraries.

**No changes are required for users who prefer buliding multiverso and 1bitsgd witih CNTK source tree** since `MULTIVERSO_PATH` and `ONEBITSGD_PATH` are set to the correct default values pointing to `Source/Multiverso` and `Source/1BitSGD` separately. I've successfully built CNTK with and without specifying the flags above.

This in-progress work targets to adding CNTK into Spack repo, a package management system originated from Supercomputer community. Feedbacks are welcome. https://github.com/LLNL/spack/pull/3578

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,14 +25,18 @@
 #     If not specified, GPU will not be enabled
 #   CUB_PATH= path to NVIDIA CUB installation, so $(CUB_PATH)/cub/cub.cuh exists
 #     defaults to /usr/local/cub-1.4.1
-#   CUDNN_PATH= path to NVIDIA cuDNN installation so $(CUDNN_PATH)/cuda/include/cudnn.h exists
+#   CUDNN_PATH= path to NVIDIA cuDNN installation so $(CUDNN_PATH)/include/cudnn.h exists
 #     CuDNN version needs to be 5.0 or higher.
 #   KALDI_PATH= Path to Kaldi
 #     If not specified, Kaldi plugins will not be built
+#   OPENFST_PATH= Path to OpenFST
+#     OpenFST is required by Kaldi Reader.
 #   OPENCV_PATH= path to OpenCV 3.1.0 installation, so $(OPENCV_PATH) exists
 #     defaults to /usr/local/opencv-3.1.0
 #   PROTOBUF_PATH= path to Protocol Buffers 3.1.0 installation, so $(PROTOBUF_PATH) exists
 #     defaults to /usr/local/protobuf-3.1.0
+#   MULTIVERSO_PATH= path to Multiverso installation or source code, so $(MULTIVERSO_PATH) exists
+#     defaults to Source/Multiverso
 #   LIBZIP_PATH= path to libzip installation, so $(LIBZIP_PATH) exists
 #     defaults to /usr/local/
 #   BOOST_PATH= path to Boost installation, so $(BOOST_PATH)/include/boost/test/unit_test.hpp
@@ -149,8 +153,8 @@ ifdef CUDA_PATH
 
 # Set up cuDNN if needed
   ifdef CUDNN_PATH
-    INCLUDEPATH += $(CUDNN_PATH)/cuda/include
-    LIBPATH += $(CUDNN_PATH)/cuda/lib64
+    INCLUDEPATH += $(CUDNN_PATH)/include
+    LIBPATH += $(CUDNN_PATH)/lib64
     LIBS_LIST += cudnn
     COMMON_FLAGS +=-DUSE_CUDNN
   endif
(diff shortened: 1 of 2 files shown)


feiga:
@weijianwen For your works about adding CNTK to Spack repo. Dose CNTK default building block your progress? Why not just build CNTK together with 1bitsgd and multiverso from scratch in your work.

I prefer to use the default source tree to build Multiverso instead of the pre-installed. One problem is that when you build Multiverso outside, how to make sure the MPI / Boost library or the build mode (Debug/Release) that multiverso uses are consistent with CNTK? We need to test to prevent any potential issues.

weijianwen:
Spack, not Spark :-) 

http://spack.readthedocs.io/en/latest/tutorial_sc16.html

My arguments are as follows. Technically Spack is able to resolve the dependency problems you concern about.

Firstly, Spack originates from HPC environment where many versions of a same library (plus various building flags) may be coexisted on the same system. Given multiple architectures (x86, PowerPC, ...) and compiler choices (GCC, PGI, ...), the resulting combination will be quite a number. Spack aims to manage this complexity by declaring fine-grain package dependency and resolve them in a DAG. Regarding your BOOST/MPI/BLAS/DEBUG concerns, Spack can guarantee component libs in CTNK will be built with consistent ones.

Secondly, building `multiverso` within `CNTK` is not the standard way to integrate external libraries. Libraries should be reused by headers and libs, not in raw source code or build process. No one includes OpenBLAS and builds it within CNTK.

Thirdly, including component's source code as building dependency may be convenient when the project initialize, but it adds increasing burden to the master project's `Makefile` since you need to duplicate every component's build process. Attempts to port CNTK to other architectures and compilers may be halted by the complexity of this Makefile. Splitting `multiverso` build away from CNTK Makefile may be a starting point to consider. 

Finally, this commit offer choices and it is not compulsory at all. Normal users just build `multiverso` within CNTK as before (**NO CHANGES REQUIRED**). Advanced users will take their own risk and try the new `--with-multiverso` and `--with-1bitsgd` flags. 

I understand CNTK project focus and priority on merging PRs, and admire community's effort to push DL to wider areas. Thus, I also appreciate your reconsideration on this commit which may server as a starting point to pave CNTK's way to the Supercomputer community. Whether this commit is accepted, I plan to patch CNTK on the Spack side and have some quick tests soon. Feedbacks from CNTK are always welcome.

veikkoeeva:
I don't want to hijack this thread, but maybe to add perspective to discussion, perhaps it [modern CMake](https://schneide.wordpress.com/2016/04/08/modern-cmake-with-target_link_libraries/) should be considered as it looks like version 3.8 handles CUDA too and the new CMake server is being integrated into a lot of tooling (such as [VS 2017](https://blogs.msdn.microsoft.com/vcblog/2017/05/10/cmake-support-in-visual-studio-whats-new-in-2017-15-3-update/)). Maybe it would become simpler to set up dependencies using strategies like [this](https://crascit.com/2015/07/25/cmake-gtest/) (downloads, and compiles Google Test automatically, works for other dependencies such as Boost too).

In addition, separately something to consider is to use (or deploy CNTK to):
- [Vspkg](https://github.com/Microsoft/vcpkg)
- [Conan C/C++ Package Manager](https://conan.io/).

I realize even thinking these would require work to be done, but maybe even recoding these points is something to consider. _Grokkability_ seems to be important in growing involvement.

JimSEOW:
@veikkoeeva pls figure out if we could simplify the CMake process so that anyone who is interested to make C# SWIG binding to be as up to date with that of python SWIG - can contribute easily. Great job!

veikkoeeva:
@JimSEOW I have not looked into CNTK build (not even tried to build the solution, I'm gearing on this still), but there's [UseSWIG](https://cmake.org/cmake/help/v3.8/module/UseSWIG.html). You probably know this already, though.

It might be useful to think it like this (mind you, I'm still very new to this): Working on a modern CMake based approach would probably be an incremental thing that takes a longer time. This likely means first it should be examined and listed what is needed and what features of CMake are the ones realizing the end goal &ndash; which may not be something the maintainers would like to have. This would give us the high-level constructs and perhaps indicate some features to start with. Overall the task would be to define smaller tasks that need to be done and coordinate them with a check-list so that we don't end up with a huge PR that burns one implementor down.

So we'd end up with a list like following:

- [x] Feature 1
- [ ] SWIG
- [ ] Feature 2

Maybe it would be feasible already now to define a small, bare-bones `CMakeLists.txt` that would define some tasks, but assume the current build does the actual job and have the [UseSWIG`](https://cmake.org/cmake/help/v3.8/module/UseSWIG.html) command doing some actual work.

JimSEOW:
@veikkoeeva Xamarin solve UrhoSharp binding from Urho3D c++ using CppCsharp than SWIG

veikkoeeva:
@JimSEOW If there isn't much API churn, P/Invoke would probably be my choice too (or first generating and then smoothign manually). There is [CNTK Gitter chat](https://gitter.im/Microsoft/CNTK) these days too. It's one place to idly discuss these things.

wolfma61:
I'm a little at a loss what is being discussed here now 
is it cmake, spack, vs2017, mutliverso/1bit, swig dependencies 

at this point I don't see the need to change the way we build multi-verso, 1bit-sgd. The way it is it guarantees that the latest bits are being picked up. This is what is getting tested and should be the default

PR #1715, opened 2017-04-07
Adding dilated convolution support

FDecaYed:
This commit has following changes:
Remove deprecated call to support cudnn v6:
Now code should compile properly with cudnn v6.

Adding dilated convolution:
Default behavior of convolution remain unchanged. Dilated convolution support is exposed through BrainScript.
User can add `dilation =` in BrainScript ConvolutionalLayer definition. Example:
`ConvolutionalLayer {96, (11:11), stride=(4:4), pad=false, dilation=1}`
`diltaion = 1` means normal convolution, which is also the default value.

Adding check on auto padding:
When (inputSize-1) is divisible by stride and filter size is even in any dimension, asymmetric padding is needed in auto padding.
Since cudnn does not support asymmetric padding, a check is added to print error in this case to avoid user getting random results.

### DIFF
--- a/Source/CNTK/BrainScript/CNTKCoreLib/CNTK.core.bs
+++ b/Source/CNTK/BrainScript/CNTKCoreLib/CNTK.core.bs
@@ -82,7 +82,7 @@ ConvolutionalLayer {numOutputChannels,   # e.g. (1) or BS.Constants.None
                     initValueScale = 1,          # TODO: rename to initScale
                     initBias = 0,
                     #reductionRank = 1,          # TODO: support this
-                    stride = 1, pad = false,
+                    stride = 1, dilation=1, pad = false,
                     lowerPad = 0, upperPad = 0,
                     maxTempMemSizeInSamples = 0} =
 {
@@ -94,7 +94,7 @@ ConvolutionalLayer {numOutputChannels,   # e.g. (1) or BS.Constants.None
     b = ParameterTensor(_ConcatArrays (Repeat (Length (filterShape), 1), outputChannelsShape), initValue = initBias)                                                       # [ 1 x 1 x     K ]
     sharing = true    # TODO: support this
     apply (x) = {
-        c = Convolution (W, x, filterShape, mapDims = numOutputChannels, stride = stride, sharing = sharing, autoPadding = pad, lowerPad = lowerPad, upperPad = upperPad, maxTempMemSizeInSamples = maxTempMemSizeInSamples)
+        c = Convolution (W, x, filterShape, mapDims = numOutputChannels, stride = stride, dilation = dilation, sharing = sharing, autoPadding = pad, lowerPad = lowerPad, upperPad = upperPad, maxTempMemSizeInSamples = maxTempMemSizeInSamples)
         res = activation (if bias then c + b else c)
     }.res
 }.apply
@@ -103,19 +103,19 @@ ConvolutionalLayer {numOutputChannels,   # e.g. (1) or BS.Constants.None
 ConvolutionTransposeLayer {numOutputChannels,
                            filterShape,         # e.g. (3:3)
                            numInputChannels,
-                           bias = true, 
-                           activation = (x=>x), 
-                           init = 'glorotUniform', 
-                           initValueScale = 0.001, 
-                           initBias = 0, 
-                           stride = 1, pad = false, 
-                           lowerPad = 0, upperPad = 0, 
-                           outputShape = None, 
+                           bias = true,
+                           activation = (x=>x),
+                           init = 'glorotUniform',
+                           initValueScale = 0.001,
+                           initBias = 0,
+                           stride = 1, pad = false,
+                           lowerPad = 0, upperPad = 0,
+                           outputShape = None,
                            maxTempMemSizeInSamples = 0} =
 {
     outputChannelsShape = _AsArray (numOutputChannels)
-    kernelShape = _ConcatArrays (filterShape, outputChannelsShape) 
-    paramShape = _ConcatArrays (kernelShape, _AsArray (numInputChannels)) 
+    kernelShape = _ConcatArrays (filterShape, outputChannelsShape)
+    paramShape = _ConcatArrays (kernelShape, _AsArray (numInputChannels))
     W = ParameterTensor{paramShape, init=init, initValueScale=initValueScale, initOnCPUOnly=true}
     b = ParameterTensor(_ConcatArrays (Repeat (Length (filterShape), 1), outputChannelsShape), initValue = initBias)
     sharing = true    # TODO: support this
@@ -151,14 +151,14 @@ AveragePoolingLayer {filterShape, stride = 1, pad = false, lowerPad = 0, upperPa
     _PoolingLayer {"average", filterShape, stride = stride, pad = pad, lowerPad = lowerPad, upperPad = upperPad, ceilOutDim = ceilOutDim, includePad = includePad}
 
 MaxUnpoolingLayer {filterShape,     # e.g. (3:3)
-                   stride = 1, 
+                   stride = 1,
                    pad = false,
-                   lowerPad = 0, 
-                   upperPad = 0} = 
+                   lowerPad = 0,
+                   upperPad = 0} =
 {
     apply (unpoolInput, poolInput) = MaxUnpooling (unpoolInput, poolInput, filterShape, stride = stride, autoPadding = pad, lowerPad = lowerPad, upperPad = upperPad)
 }.apply
-    
+
 # RecurrentLSTMLayer -- create an LSTM layer
 RecurrentLSTMLayer {outputDim,
                     cellShape = None, # if set then use a projection
@@ -393,10 +393,10 @@ _AsNodes (inputs) = {
 }.arr
 
 ##############################################################################
-# "Stable API" with the purpose of staying compatible towards 2.0. 
+# "Stable API" with the purpose of staying compatible towards 2.0.
 # - Use only tensors as concept. Move away from matrices.
 # - Main input goes first
-# - Main input is called "_" (becomes either self, or is moved to the end, 
+# - Main input is called "_" (becomes either self, or is moved to the end,
 #   depending on language binding)
 # - tensor shape is called "shape"
 # - output shape before input shape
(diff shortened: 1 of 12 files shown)


FDecaYed:
This code will also compile without issue with cudnn v5. Though it will error out if user uses dilation > 1.

FDecaYed:
A lot happened since this PR issued. Just updated it as follow:
Things like v6 support, error check are already in master now, thus removed.
Now it only contains support to use dilated convolution with cudnn 6 in Brainscript.

I will be working on python binding and a unit test in python next.

FDecaYed:
PR updated with python API for dilated convolution and unit test.

JohannesKummerer:
When i first compiled it, VS dropped a Warning:
"C4100 'dilation': unreferenced formal parameter MathCUDA 
c:\repos2\cntk\cntk\source\math\ConvolveGeometry.h 491"
and did not compile.
I tried to ignore it with the macro:
 "UNREFERENCED_PARAMETER(dilation);" in ConvolveGeometry.h  line 493.
After that change, it compiled. But i could only use a dilation=1(no dilation).
If i try dilation=2, cntk fails: 
"cuDNN failure 8: CUDNN_STATUS_EXECUTION_FAILED ; GPU=0 ; hostname=TUEIMMK-DAN10 ; expr=workspaceSizeFinder()"

I use VS2015Update3, CUDAv8.0, cuDnn-8.0-v6.0, cub-1.4.1, Windows 10, GTX970

FDecaYed:
@JohannesKummerer Sorry for not seeing your comment, been busy this month.
The compile error you see is because that I did not implement calculation of size considering dilation in computeinputshape. That function is used in deconvolution(transposed conv), and if you are not using deconvolution, that compile error should not affect your run.
And if you are not using deconv, CUDNN_STATUS_EXECUTION_FAILED could mean a lot of things. I may need more information or a repro to help you on this.

KeDengMS:
Hi @FDecaYed, now that CNTK has moved to CuDNN 6, can you resolve the conflicts and update the PR?

FDecaYed:
@KeDengMS Hi Ke,
I just rebased the branch on to master. There is a kernel_test in bindings/python/cntk/ops/tests/kernel_test.py. --device=gpu should pass and running on cpu will give you runtime error.

KeDengMS:
Thanks @FDecaYed. I understand that dilated convolution is not support in CPU yet, can you skip the test for CPU similar to [this](https://github.com/Microsoft/CNTK/blob/master/bindings/python/cntk/ops/tests/function_tests.py#L349-L353)?

FDecaYed:
Just updated it to resolve conflict after recent cudnn conv engine fix

Source/CNTKv2LibraryDll/API/CNTKLibrary.h
```diff
@@ -2,7 +2,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE.md file in the project root for full license information.
 //
-// This is the main header of the CNTK library API containing the entire public API definition. 
+// This is the main header of the CNTK library API containing the entire public API definition.
```
ebarsoumMS: Why all the carriage return and tab change?

Source/CNTKv2LibraryDll/API/CNTKLibrary.h
```diff
@@ -4142,27 +4142,29 @@ namespace CNTK
...
-        const std::vector<bool>& sharing = { true },
-        const std::vector<bool>& autoPadding = { true },
-        const NDShape& outputShape = { 0 },
-        size_t maxTempMemSizeInSamples = 0,
-        const std::wstring& name = L"");
+                                              const Variable& operand,
+                                              const NDShape& strides = { 1 },
+                                              const std::vector<bool>& sharing = { true },
+                                              const std::vector<bool>& autoPadding = { true },
+                                              const NDShape& dilation = { 1 },
```
ebarsoumMS: dilation should be after outputshape, don't break compatibility.

Source/CNTK/BrainScript/CNTKCoreLib/CNTK.core.bs
```diff
@@ -82,7 +82,7 @@ ConvolutionalLayer {numOutputChannels,   # e.g. (1) or BS.Constants.None
                     initValueScale = 1,          # TODO: rename to initScale
                     initBias = 0,
                     #reductionRank = 1,          # TODO: support this
-                    stride = 1, pad = false,
+                    stride = 1, dilation=1, pad = false,
```
ebarsoumMS: That will break compatibility, put dilation before maxTemMemSize... only.

Source/Math/ConvolveGeometry.h
```diff
@@ -461,8 +470,7 @@ class ConvolveGeometry final
             if (mapCount.size() > 1)
                 dimOut *= mapCount[i];
             else if (i == inputShape.GetRank() - 1)
-                dimOut *= mapCount[0];
-
+                dimOut = dimOutNoDilation*mapCount[0];
```
ebarsoumMS: Are you sure this part is correct? If there is no dilation, should dimOut *=mapCount[0]?

Source/Math/ConvolveGeometry.h
```diff
@@ -461,8 +470,7 @@ class ConvolveGeometry final
             if (mapCount.size() > 1)
                 dimOut *= mapCount[i];
```
ebarsoumMS: Does dilation affect this line?

Source/CNTKv2LibraryDll/API/CNTKLibrary.h
FDecaYed: My editor did this. Is there any reason for those tailing space/tab in the first place?

ebarsoumMS: No, but changing them as part of this review, make it a little difficult to cr the change.

Source/Math/ConvolveGeometry.h
FDecaYed: My understanding:
this whole '**dimout *='** thing is a hacky way to make the dimension calculation work for both convolution and pooling. It is handcrafted in a way that **dimOut** is both the output size for some dimension, and used as indicator for conv/pooling and stored in some place unoccupied.
My changes did the right culculation in the 'real dimension' with dilation but break the formula for the 'indicator' dimension.
So for simplicity I used the old formula without dilation to calculate the old value and put there. **dimOutNoDilation** means use old formula without considering dilation.

Source/CNTK/BrainScript/CNTKCoreLib/CNTK.core.bs
FDecaYed: Please make any necessary change to keep compatibility, it could be easier on your side with more test available.
This code has been in side branch for a long time and I could have introduced some compatibility issue merging upstream change over time.

Source/CNTKv2LibraryDll/API/CNTKLibrary.h
FDecaYed: We can't run full EndToEnd test on our side due to data dependency. There might be issue I did not catch.

Source/CNTKv2LibraryDll/API/CNTKLibrary.h
FDecaYed: Sorry about the inconvenience. I just realize github does not have hide whitespace function.

Source/CNTK/BrainScript/CNTKCoreLib/CNTK.core.bs
ebarsoumMS: Can you atleast rebase it with the latest form master?

FDecaYed: @ebarsoumMS Just squashed all commit and rebased onto master

ebarsoumMS:
Can you resolve the conflict?

FDecaYed:
@ebarsoumMS Rebased onto master

ebarsoumMS:
@FDecaYed even python unit test is broken, those test don't need extra dependency, just run pytest on binding/python folder.

If you run the below code, it will break:

import cntk as C
import numpy as np

nC, inH, inW = 2, 6, 7
y = C.input_variable((inC, inH, inW))
in_filter_shape = (3, 2)
out_num_filters = 4

zeropad = False
in_strides = 1

model = C.layers.Convolution(in_filter_shape,
                             num_filters=out_num_filters,
                             activation=None,
                             pad=zeropad,
                             strides=in_strides, name='foo')
model_shape = model(y).foo.shape

The output shape is wrong.

Source/Math/ConvolveGeometry.h
```diff
@@ -440,16 +446,19 @@ class ConvolveGeometry final
...
             {
-                dim += kernelShape[i] - 1;
+                dim += dil * (kernelShape[i] - 1);
             }
             else
             {
                 dim += lo + hi;
             }
-            float preciseDimOut = (float)(dim - kernelShape[i]) / delta + 1;
+            float preciseDimOut = (float)(dim - (kernelShape[i]-1) * dil - 1) / delta + 1;
```
ebarsoumMS: Shouldn't this "(kernelShape[i]-1) * dil - 1" be "(kernelShape[i]-1) * dil + 1"? If kernel is 5 and dil is 2, then you want 9 not 7, right? Also, with this change you don't need something special for dilation=1.

Source/CNTKv2LibraryDll/PrimitiveFunction.h
```diff
@@ -713,8 +714,8 @@ namespace CNTK
         static void FixNDShape(size_t filterRank, size_t inputRank, NDShape& shape, size_t deflt, const NDShape& from = NDShape());
 
         static NDShape ConvolutionOpOutputShape(PrimitiveOpType op, const NDShape& operandShape, NDShape& kernelShape, NDShape& outputMapCount, NDShape& strides,
-            std::vector<bool>& sharing, std::vector<bool>& autoPad, NDShape& lowerPad, NDShape& upperPad,
-            bool transpose, bool inferDimensions, bool ceilOutputDim = false);
+                                                std::vector<bool>& sharing, std::vector<bool>& autoPad, NDShape& lowerPad, NDShape& upperPad,
+                                                bool transpose, bool inferDimensions, NDShape dilation = NDShape({1}), bool ceilOutputDim = false);
```
ebarsoumMS: NDShape dilation should be by reference, NDShape& dilation.

ebarsoumMS: Also FixNDShape(fillRank, inputRank, dilation, 1); instead of FixNDShape(fillRank, inputRank, dilation, 1, fromShape);. The latter will add non 1 at the end.

Source/Math/ConvolveGeometry.h
FDecaYed: I agree with you. But isn't "dim-[(kernelShape[i]-1) * dil + 1]" is the same as "dim-(kernelShape[i]-1) * dil - 1" here? Which is exactly what I did.

ebarsoumMS: Sorry, my bad. But with that, you don't need the other variabkle with NoDilation on them, because when dil is 1, it fall back to the right one.

Here my update:

            size_t dil = dilation[dilation.GetRank() == 1 ? 0 : i];
            if (autoPadCur)
            {
                dim += dil * (kernelShape[i] - 1);
            }
            else
            {
                dim += lo + hi;
            }

            size_t effectiveKernelShape = (kernelShape[i] - 1) * dil + 1;
            float preciseDimOut = (float)(dim - effectiveKernelShape) / delta + 1;
            size_t dimOut = static_cast<size_t>(ceilOutDim ? ceil(preciseDimOut) : floor(preciseDimOut));
            // When LowerPad and/or UpperPad are specified (i.e. > 0), we insist that the kernel applications
            // fill the entire space.
            if (!autoPadCur && (lo > 0 || hi > 0))
            {
                size_t size = (dimOut - 1) * delta + kernelShape[i];
                if (size != dim)
                    InvalidArgument("Convolution requires that kernel fills the entire space if auto-padding is disabled.");
            }

            if (mapCount.size() > 1)
                dimOut *= mapCount[i];
            else if (i == inputShape.GetRank() - 1)
                dimOut *= mapCount[0];

            dimsOutput[i] = dimOut;

FDecaYed: If you combine this piece of code with your approach of set last dimension of dilation to 1, it should work the same as what I did here.
I basically want treat last dimension differently as if there is no dilation(=1).

The problem is that I don't understand 2 problems:
1. why my code generate [1,1,2] from 1, this was definitely not intended.
2. my code should generate [2,2,2] from 2, but because of above code, the last dimension should be handled specially, so the number in last dimension should have no effect anyway.

ebarsoumMS: 1. For the [1,1,2] because the fromShape had 2 at the end. 
2. I can make it to do that, what about dilation of 3, should it expand to [3,3,3]? WHat if I have [3,2], how should we expand the channel axis.

PR #1790, opened 2017-04-23
Add Nadam method

fykjia:


KeDengMS:
This change conflicts with master, and adding a learner should not involve so many files. I'll close this PR for now. Please submit another up-to-date PR with only the files that need to change.

PR #1812, opened 2017-04-28
Add adamax option for Adam learner.

walkerning:
Add an option `adamax` for Adam Learner, which uses inifinity-norm for variance momemtum running average.

For a simple test, in ["ResNet20 on CIFAR10"](https://github.com/Microsoft/CNTK/blob/master/Examples/Image/Classification/ResNet/Python/TrainResNet_CIFAR10.py) tutorials，I compare the results of the adamax optimizer and of the original adam optimizer. 

The configuration of hyper parameters remain the same as in the original tutorial.
```
lr_per_minibatch       = learning_rate_schedule([1.0]*80 + [0.1]*40 + [0.01], UnitType.minibatch, epoch_size)
gradient_momentum =  momentum_as_time_constant_schedule(-minibatch_size/np.log(0.9))
```
Variance momentum is remained as the default value.

To use inifity-norm in the adam learner, just add an option `adamax=True`:
```
learner = adam(z.parameters, lr_per_minibatch, momentum_time_constant, l2_regularization_weight=l2_reg_weight, adamax=True)
```

Run the training process for 200 epochs, the training-set error rate/loss results is shown below.
image

On the test set, the error rate of Adam and Adamax is 9.4% and 8.8% respectively.

I'm now running more tests

### DIFF
--- a/Source/CNTKv2LibraryDll/API/CNTKLibrary.h
+++ b/Source/CNTKv2LibraryDll/API/CNTKLibrary.h
@@ -4455,6 +4455,7 @@ namespace CNTK
                                     bool unitGain = DefaultUnitGainValue(),
                                     const MomentumSchedule& varianceMomentumSchedule = DefaultVarianceMomentum,
                                     double epsilon = 1e-8,
+                                    bool adamax = false,
                                     AdditionalLearningOptions additionalOptions = AdditionalLearningOptions());
 
     ///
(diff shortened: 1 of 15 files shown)


KeDengMS:
Thanks for the contribution @walkerning. Can you rebase to latest master?

walkerning:
@KeDengMS Done~

PR #1817, opened 2017-04-30
A simple implementation of Adamax

Gushan33:
A simple implementation of Adamax


KeDengMS:
Adamax is available now as an option to Adam. https://github.com/Microsoft/CNTK/pull/1812

PR #1820, opened 2017-04-30
ResNet example to compare different optimizers. 

songzy12:
Here we add an example code to run four different optimizers on the Image Classification, and plot the differences between loss, metric, time using matplot. Also, AdaDelta Optimizer is implemented and the comparison is also added as above.

### DIFF
--- a/Manual/Manual_How_to_use_and_evaluate_different_learners.ipynb
+++ b/Manual/Manual_How_to_use_and_evaluate_different_learners.ipynb
@@ -1,0 +1,649 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In this tutorial, we will show how to use the [code](https://github.com/Microsoft/CNTK/blob/master/Examples/Image/Classification/ResNet/Python/TrainResNet_CIFAR10.py) in the Examples directory to evaluate the performance (i.e., time cost, loss, metric) of different learners (i.e., sgd, momentum_sgd, nesterov, adam) on the classification task of CIFAR-10 dataset using ResNet model.\n",
+    "\n",
+    "The structure of this tutorial are organized as follows: we first introduce the CIFAR-10 dataset, ResNet model, and different learners. Then we will show how to use these learners during the training of ResNet model on the CIFAR-10 dataset, after which we will plot the time cost, loss, and metric trends by time."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## CIFAR-10\n",
+    "\n",
+    "### Dataset\n",
+    "\n",
+    "The CIFAR-10 dataset [2] contains 6k 32×32 color images for each of 10\n",
+    "classes, therefore there are 60k images in total. \n",
+    "\n",
+    "The 10 classes are: airplane, automobile, bird, cat, deer, dog, frog, horse, ship, truck.\n",
+    "\n",
+    "Note that these classes are mutual exclusive.\n",
+    "\n",
+    "The whole set is further divided into 5 training sets and 1 test set, with 10k\n",
+    "images for each. \n",
+    "\n",
+    "The test set contains exactly 1k for each class, while this condition\n",
+    "is not necessary for the training sets.\n",
+    "\n",
+    "### Download\n",
+    "\n",
+    "We can use this [script](https://github.com/Microsoft/CNTK/tree/master/Examples/Image/DataSets/CIFAR-10/install_cifar10.py)  to download the dataset. Simple run the following command in a terminal:\n",
+    "\n",
+    "```bash\n",
+    "cd Examples/Image/DataSets/CIFAR-10\n",
+    "python install_cifar10.py\n",
+    "```\n",
+    "\n",
+    "**Note**: Error is encountered when I run this script on my Ubuntu 17.04 64 bit machine, while it runs well on my Win10 64bit machine.\n",
+    "\n",
+    "The tree structure of the whole dataset is as follows:\n",
+    "* `cifar-10-batches-py` includes 5 training batches and one test batch.\n",
+    "* `train` and `test`: include the 60k `.png` files.\n",
+    "* `train_map.txt` and `test_map.txt` maintain the location of each `.png` image and corresponding label.\n",
+    "* `Train_cntk_text.txt` and `Test_cntk_text.txt` have a label and corresponding features (pixel level, 32 * 32 * 3) for one image at each line."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "C:\\repos\\CNTK\\Manual\\..\\Examples\\Image\\DataSets\\CIFAR-10\n"
+     ]
+    }
+   ],
+   "source": [
+    "import os\n",
+    "\n",
+    "# Paths relative to current python file.\n",
+    "abs_path = os.path.dirname(os.path.abspath('__file__'))\n",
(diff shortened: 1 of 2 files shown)


KeDengMS:
Hi @songzy12, thanks for the contribution to CNTK! I believe there will be other users interested in the comparison on different learners.
One thing I'd like to point out is that we already have built-in adadelta learner, and it might be faster than UserLearner version.
Besides, it seems TrainResNet_CIFAR10.py is a bit overloaded than its original intention. We also have test cases built for it which might be affected with your change. Is it better to turn your code into a jupyter notebook under Tutorial? That would allow others to interactively change the script as well.

songzy12:
@KeDengMS Thank you for your reply!
When I decided to implement the UserLearner version adadelta , I was refering to [a older version of learners](https://github.com/Microsoft/CNTK/blob/8e3a398d9a51c04a82ed5bbb376623dc881bbe8a/bindings/python/cntk/learners/__init__.py), which is not the intended one. My bad.
In the following couple of days I will try to turn the comparison part into a jupyter notebook. Hope that can be done quickly. :)

songzy12:
@KeDengMS I wrote something in the jupyter notebook about what I modified to run the experiment. That is the first time I did something like this, so I wonder whether it is suitable for a merge?

KeDengMS:
@songzy12 Thanks for the update! We'd love to see contributions from community to make CNTK better. Please keep only the notebook part in tutorial, and use the built-in adadelta learner in the notebook. If you want to show the usage of user learner please consider implementing another one that's not built in. @sayanpa may give you more feedbacks on how to get it in as a tutorial on choosing learners. Sorry that our responses might be slow while we are working towards a milestone.

Examples/Image/Classification/ResNet/Python/TrainResNet_CIFAR10.py
```diff
@@ -7,18 +7,48 @@
 from __future__ import print_function
 import os
 import argparse
+import code
```
KeDengMS: This file should not be modified as it may cause some other test changes. Only notebook is needed.

Examples/Image/Classification/ResNet/Python/.gitignore
```diff
@@ -0,0 +1 @@
+Models/
```
KeDengMS: This file should not change

Examples/Image/Classification/ResNet/Python/TrainResNet_CIFAR10.py
```diff
@@ -7,18 +7,48 @@
...
 from cntk import Trainer, cntk_py
-from cntk.learners import momentum_sgd, learning_rate_schedule, momentum_as_time_constant_schedule, UnitType
+from cntk.learners import adam, nesterov, sgd, momentum_sgd, learning_rate_schedule, momentum_as_time_constant_schedule, UnitType
+from cntk.learners import UserLearner
 from cntk.debugging import set_computation_network_trace_level
 from cntk.logging import *
 from cntk.debugging import *
 from resnet_models import *
 
+class AdaDelta(UserLearner):
```
KeDengMS: There is built-in adadelta learner already.

Examples/Image/Classification/ResNet/Python/script/Benchmark.py
```diff
@@ -0,0 +1,61 @@
+import re
+import code
```
KeDengMS: Should be part of the notebook, not ResNet example

Tutorials/CNTK_201C_CIFAR-10_ResNet20_Learner_Comparison.ipynb
```diff
@@ -0,0 +1,363 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "from cntk.learners import UserLearner\n",
+    "\n",
+    "class AdaDelta(UserLearner):\n",
```
KeDengMS: Please use built-in adadelta

Examples/Image/Classification/ResNet/Python/script/Benchmark.py
sayanpa: Assuming these plots can be generated as a part of the notebook

Tutorials/CNTK_201C_CIFAR-10_ResNet20_Learner_Comparison.ipynb
sayanpa: I would like this to move under CNTK/Manual. Make the title Manual_how_to_use_and_evaluate_different_learners.ipynb.

Please also add a test for this notebook under CNTK/Tests/EndToEndTests/CNTKv2Python/Manual

KeDengMS:
Thanks for merging with master. Can you make changes as Sayan suggested?

1. Move under CNTK/Manual. Make the title Manual_how_to_use_and_evaluate_different_learners.ipynb.
2. Please also add a test for this notebook under CNTK/Tests/EndToEndTests/CNTKv2Python/Manual

Besides, please squash the commits and avoid unnecessary file changes for the cleanness of log history.

songzy12:
Hello, how about now?

KeDengMS:
Thanks for the update. However when I was trying to merged it through CI system I hit following failures:
```
14:42:24 CNTKv2Python/Manual: nbconvert.preprocessors.execute.CellExecutionError: An error occurred while executing the following cell:
14:42:24 CNTKv2Python/Manual: ------------------
14:42:24 CNTKv2Python/Manual: log_dir='log/'
14:42:24 CNTKv2Python/Manual: 
14:42:24 CNTKv2Python/Manual: if not os.path.exists(log_dir):
14:42:24 CNTKv2Python/Manual:     os.makedirs(log_dir)
14:42:24 CNTKv2Python/Manual: 
14:42:24 CNTKv2Python/Manual: learners = ['adam', 'momentum_sgd', 'nesterov', 'sgd']
14:42:24 CNTKv2Python/Manual: for learner in learners:
14:42:24 CNTKv2Python/Manual:     reader_train = create_image_mb_source(os.path.join(data_path, 'train_map.txt'), os.path.join(data_path, 'CIFAR-10_mean.xml'), True, total_number_of_samples=epochs * epoch_size)
14:42:24 CNTKv2Python/Manual:     reader_test = create_image_mb_source(os.path.join(data_path, 'test_map.txt'), os.path.join(data_path, 'CIFAR-10_mean.xml'), False, total_number_of_samples=C.io.FULL_DATA_SWEEP)
14:42:24 CNTKv2Python/Manual:     train_and_evaluate(reader_train, reader_test, network_name, epoch_size, epochs, learner=learner, model_dir=model_dir, log_dir=log_dir)
14:42:24 CNTKv2Python/Manual: ------------------
14:42:24 CNTKv2Python/Manual: 
14:42:24 CNTKv2Python/Manual: RuntimeError: File '/home/ubuntu/workspace/Examples/Image/Classification/ResNet/Python/../../../DataSets/CIFAR-10/train_map.txt' or '/home/ubuntu/workspace/Examples/Image/Classification/ResNet/Python/../../../DataSets/CIFAR-10/CIFAR-10_mean.xml' does not exist. Please run install_cifar10.py from DataSets/CIFAR-10 to fetch them
14:42:24 CNTKv2Python/Manual: 
```
Seems to be a data path issue. Please take a look at other tutorials/manuals on how dataset is specified with test.

songzy12:
@KeDengMS  The reason this error has appeared is that the data set is not installed by default. The section 1 in this Manual describes how to install this data set by hand (where a command `python install_cifar10.py` need to be run in `cmd` under some directory).

The author of [Manual_How_to_feed_data](https://github.com/Microsoft/CNTK/blob/master/Manual/Manual_How_to_feed_data.ipynb) decides to raise an exception when the data set is missing, but I decide to integrate the download python code into this manual to make it self-contained. 

Now it should be fine.

songzy12:
That may be due to some previous installed files.

I add some exception handling code for this part.. I hope it is fine now..

PR #1825, opened 2017-04-30
Add a new optimization method:Adamax

greathall:
A new optimization method:Adamax for Resnet_cifar10


KeDengMS:
Adamax is available now as an option to Adam. https://github.com/Microsoft/CNTK/pull/1812

PR #1826, opened 2017-04-30
Adamax Learner

swnhieian:
Implementation of a user-defined Learner using python api, with example usage in TrainResNet_CIFAR10.py


KeDengMS:
A C++ implementation of Adamax has been merged. https://github.com/Microsoft/CNTK/pull/1812

PR #1853, opened 2017-05-07
implemented proximal gradient descent learner

jackjlynch:


Ark-kun:
Can you please explain the difference between this learner a the default sgd learner?

PR #1854, opened 2017-05-07
Test an internal optimization method: SGD

neutrinogxy:
Change Momentum SGD to SGD with original l2 regularization.


KeDengMS:
Seems to be user experimental changes only. Closing for now.

Issue #1928, opened 2017-05-30
ProgressPrinter Heartbeat


vignesh-m:
Currently, when the gen_heartbeat parameter is set to True in a ProgressPrinter, all that is printed is PROGRESS : 0.00%. 

This happens when Examples/Image/Classification/ConvNet/Python/ConvNet_CIFAR10_DataAug_Distributed.py is executed, for example.

This is caused by the code in [progress_print.py:250](https://github.com/Microsoft/CNTK/blob/master/bindings/python/cntk/logging/progress_print.py#L250)

Since the ProgressPrinter.update method is deprecated, trainers do not update the self.epochs variable. One solution I could make work was to set self.epochs when [on_write_training_summary](https://github.com/Microsoft/CNTK/blob/master/bindings/python/cntk/logging/progress_print.py#L391) is called. Then, we can use this and the self.num_epochs to give an epoch progress.

This can be a great improvement as running this on a platform like philly will give accurate progress rather than just 0%.

cha-zhang:
Sure. Could you please create a pull request for this? Thanks!

vignesh-m:
I have created the pull request #1929

cha-zhang:
Philly no longer need progress print. Thus we will take this issue as resolved. Thanks!

PR #1929, opened 2017-05-30
Better heartbeat for ProgressPrinter. See Issue #1928

vignesh-m:
Somewhat quickfix solution outlined in Issue #1928

### DIFF
--- a/bindings/python/cntk/logging/progress_print.py
+++ b/bindings/python/cntk/logging/progress_print.py
@@ -252,7 +252,7 @@ class ProgressPrinter(cntk_py.ProgressWriter):
         # print progress no sooner than 10s apart
         if timer_delta > 10 and self.gen_heartbeat:
             # print to stdout
-            print("PROGRESS: 0.00%")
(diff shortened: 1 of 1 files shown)


bindings/python/cntk/logging/progress_print.py
```diff
@@ -398,6 +398,7 @@ def on_write_training_summary(self, samples, updates, summaries, aggregate_loss,
         elapsed_seconds = elapsed_milliseconds / 1000
         speed = _avg(samples, elapsed_seconds)
         avg_loss = _avg(aggregate_loss, samples)
+        self.epochs = summaries
```
eldakms: I suppose this won't work in case of checkpoint/crash/restore, but still better than 0

vignesh-m: Yeah, I meant this as a quick fix to give some sense of progress

eldakms:
Please check "summaries" - it is of type string in some cases (make sure CNTKv2Python/Examples pass). Could you also please rebase on the latest master? Thank you!

cha-zhang:
Philly no longer need heartbeat signal. Close this now.

PR #1940, opened 2017-06-01
Converted MPICommunicatorImpl to singleton.

nluehr:
Avoids repeated initalization of NCCL in evaluator's test methods.


junjieqian:
@eldak is added to the review. #Closed

junjieqian:
LGTM
One suggestion from Amit, 

> The only bit of feedback I have is a request to add a regression test for the issue that this change addresses

PR #1943, opened 2017-06-01
Fix typo for Animals dataset description

juice500ml:


cha-zhang:
Thank you for the fix!

PR #1955, opened 2017-06-03
Bugfix for get_data_type

boeddeker:
get_data_type returns the wrong dtype for np.float64(1).

Example code:
``` python
import cntk
import numpy as np
for a in [
    np.float64(1),
    np.float32(1),
    np.int64(1),
    1,
    1.,
]:
    print(isinstance(a, np.inexact), cntk.utils.get_data_type(a))


```
Old output:
```
True <class 'numpy.float32'>
True <class 'numpy.float32'>
False <class 'numpy.float32'>
False <class 'numpy.float32'>
False <class 'numpy.float32'>
```

New output:
```
True <class 'numpy.float64'>
True <class 'numpy.float32'>
False <class 'numpy.float32'>
False <class 'numpy.float32'>
False <class 'numpy.float32'>
```


cha-zhang:
Can you add a test for this?

boeddeker:
Sure, I added the examples from my first comment as tests.

PR #1971, opened 2017-06-06
Bug: You can not borrow a temporary array.

boeddeker:
When np.ascontiguousarray is called a temporary array is created. This will be deleted after the function and cntk can not borrow the array without holding the object.

Here some example code to produce the bug (may need more the one execution):
```
import cntk
import numpy as np

np_array = np.ones((6, 400, 513))
y = cntk.cntk_py.NDArrayView(np_array, cntk.device.use_default_device(), False, True)

del np_array

from time import sleep
sleep(0.05)

print(y.to_ndarray())
```

The error happen, when the input for a cntk function is not c contiguous.

### DIFF
--- a/bindings/python/cntk/core.py
+++ b/bindings/python/cntk/core.py
@@ -35,7 +35,8 @@ class NDArrayView(cntk_py.NDArrayView):
         data_type (np.float32, np.float64): data type of the data
         device (:class:`~cntk.device.DeviceDescriptor`): device this value
          should be put on
-    '''
+
+'''
 
     def __init__(self, shape, data_type, device=None):
         from cntk.internal import sanitize_shape, sanitize_dtype_cntk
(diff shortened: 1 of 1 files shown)


boeddeker:
I have a better idea to fix this bug.
Store the numpy array as a property of the view. In this way it is guarantied, that the live time of the array is long enough and there is no copy overhead.

```python
import numpy as np
from cntk.core import NDArrayView
x0 = np.ones(shape=(1000))
v = NDArrayView.from_dense(x0, borrow=True)
v.ndarray = x0  # store the numpy array as property of the view. 
del x0
import time
time.sleep(0.1)
v.asarray().tolist()  # always a one array
```

```python
x0 = np.ones(shape=(1000))
v = NDArrayView.from_dense(x0, borrow=True)
# v.ndarray = x0
del x0
import time
time.sleep(0.1)
v.asarray().tolist()  # the first values are always wrong
```

PR #1977, opened 2017-06-07
Add distributed version of TransferLearning

bebosudo:
I wrote a distributed version of the TransferLearning script, by following these instructions: [wiki/Multiple-GPUs-and-machines#2-configuring-parallel-training-in-cntk-in-python](https://github.com/Microsoft/CNTK/wiki/Multiple-GPUs-and-machines#2-configuring-parallel-training-in-cntk-in-python)

Moreover, I integrated argparse in it, similarly to what is already done here: https://github.com/Microsoft/CNTK/blob/master/Examples/Image/Classification/ResNet/Python/TrainResNet_CIFAR10_Distributed.py

The script can be run in the same way as the ones already present in the same folder /Example/Image/TransferLearning/ and described in the wiki. The wiki could be updated by adding the following two lines of instructions:
```
$ printf "['wolf', 'sheep']" > labels.txt
$ mpirun -n $(ls /dev/nvidia* |grep nvidia[0-9][0-9]* |wc -l) python TransferLearning_distributed.py --train-map-file ../DataSets/Animals/Train/map.txt --test-map-file ../DataSets/Animals/Test/map.txt --labels-list-file labels.txt --batch-size 100 --epochs=30
```
The mpirun part is needed in order to start multiple processes running the distributed training. The bash expansion counts the number of GPUs allocated on the node and spawns that same number of processes.

The script can also be used just to evaluate an already trained model, by passing it the model to be evaluated and the test map file with the option --test-map-file (no mpirun is needed when evaluating).
```
$ python TransferLearning_distributed.py --evaluate-only --test-map-file ../DataSets/Animals/Test/map.txt --model-file ../TransferLearning/Output/already_trained.model
```

The whole list of options can be retrieved with:
```
$ python TransferLearning_distributed.py --help
```

I haven't managed to test whether this script works on multiple nodes (which should be defined e.g. with a hostfile passed to mpirun) because I'm working behind docker containers, and there are some problems related to networking, but theoretically it should work (given that ssh daemons are listening and that password-less login is working).

I would like to extend it further, by e.g. supporting multiple neural networks; if you have suggestions about possible edits feel free to give any advice.

### DIFF
--- a/Examples/Image/TransferLearning/TransferLearning_distributed.py
+++ b/Examples/Image/TransferLearning/TransferLearning_distributed.py
@@ -1,0 +1,266 @@
+# Copyright (c) Microsoft. All rights reserved.
+
+# Licensed under the MIT license. See LICENSE.md file in the project root
+# for full license information.
+# ==============================================================================
+
+from __future__ import print_function
+import numpy as np
+import os
+import sys
+import time
+import argparse
+from TransferLearning import (create_mb_source, create_model,
+                              eval_single_image,
+                              mb_size, lr_per_mb, momentum_per_mb,
+                              l2_reg_weight,
+                              features_stream_name, label_stream_name)
+from cntk import distributed
+from PIL import Image
+from cntk.device import try_set_default_device, gpu
+from cntk import load_model, placeholder, Constant
+from cntk import Trainer, UnitType
+from cntk.logging.graph import find_by_name, get_node_outputs
+from cntk.io import MinibatchSource, ImageDeserializer, StreamDefs, StreamDef
+import cntk.io.transforms as xforms
+from cntk.layers import Dense
+from cntk.learners import momentum_sgd, learning_rate_schedule, momentum_schedule
+from cntk.ops import input, combine, softmax
+from cntk.ops.functions import CloneMethod
+from cntk.losses import cross_entropy_with_softmax
+from cntk.metrics import classification_error
+from cntk.logging import log_number_of_parameters, ProgressPrinter
+
+
+# Define base model location and characteristics.
+base_folder = os.path.dirname(os.path.abspath(__file__))
+tl_model_file = os.path.join(
+    base_folder, "..", "PretrainedModels", "ResNet_18.model")
+now = time.strftime("%F_%H-%M-%S")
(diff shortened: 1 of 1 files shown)


bebosudo:
I'd like to contribute and see the PR done, but I don't get how to sign/agree with the microsoft CLA and I haven't received any notification/email.
Any hint?

Examples/Image/TransferLearning/TransferLearning_distributed.py
```diff
@@ -0,0 +1,275 @@
...
+                        help="Disable output (errors are still printed)",
+                        action="store_true", required=False, default=False)
+
+    args = vars(parser.parse_args())
+    VERBOSE = not args["quiet"]
+
+
+    if not (args.get("train_map_file", False) or 
+            args.get("evaluate_only", False)):
+        print("\n ERROR: Choose whether to transfer learning on a new model "
```
pkranen: ... to train a new transfer learning model ...

Examples/Image/TransferLearning/TransferLearning_distributed.py
```diff
@@ -0,0 +1,275 @@
...
+    if distributed.Communicator.rank() == 0:
+
+        if args["evaluate_only"] and args["model_file"] == tl_model_file:
+            print("\n ERROR: Please provide a trained model to be used to "
+                  "evaluate, we cannot use the default model: '{}'.".format(
+                        tl_model_file
+                    ),
+                  file=sys.stderr)
+            exit(1)
+        elif args["evaluate_only"]:
```
pkranen: Consider to change this to a nested "if args["model_file"] == tl_model_file: ... else: ..."

Examples/Image/TransferLearning/TransferLearning_distributed.py
```diff
@@ -0,0 +1,275 @@
+# Copyright (c) Microsoft. All rights reserved.
+
```
pkranen: Thanks for contributing!

Please add a test for this new script. See here (https://github.com/Microsoft/CNTK/blob/master/Tests/EndToEndTests/CNTKv2Python/Examples/transfer_learning_test.py) as an example.

PR #1996, opened 2017-06-13
Fix GCC compilation issues

bfredl:
gcc interprets this as valQ.Quantize operator< ZeroThresholdFor1Bit


bfredl:
I needed one more change to compile with gcc: make a templated member function inline.

PR #2072, opened 2017-07-02
implement selu activation function

lakshayg:
This PR implements the SELU activation function as described in https://arxiv.org/abs/1706.02515.


lakshayg:
@ebarsoumMS Thanks for the feedback. I have made the changes you suggested. Please review.

jeandebleau:
I tested the two versions of your implementation. The last one is more compact but consumes much more memory that the first version. Could you verify that ?

lakshayg:
It would be helpful if you could post the code you used for benchmarking (I haven't used CNTK before).

lakshayg:
On a second thought, it seems likely that the second version will consume much more memory. In the first version, SELU was a single computation node but after the revision, it creates several other nodes like `Less`, `ElementSelect` etc for computation.

ebarsoumMS:
I am running the above code in our Jenkins, once it pass all tests, it should be checked in automatic. @jeandebleau can you share the code for the memory issue?

jeandebleau:
I could share my code but the easiest way would be to replace the activation function of an existing cntk example with the new selu one. Of course, the example may not work anymore because of some weight initialisation issue but you could track the memory usage this way. I tested it in python using the current master version of cntk.

By the way I think some similar issues can probably be found in other activation functions such as LeakyReLU, or PReLU.

 I do not know if the creation of intermediate variables affects the memory and what is the effect on the gradient computation during learning. I guess the second option will be slower and wiil take more memory since the gradient is not explicitely defined.

PR #2078, opened 2017-07-04
'totalSamplesSeen' cast as Int64 instead of Int32 to avoid overflow

vermorel:
When performing long runtime training tasks, the
number `totalSamplesSeen` as printed by `cntk.exe`
does overflow at 2^31, and the console starts displaying
negative counts (situation actually observed). The
overflow is caused by a cast to `Int32` which can be
upgraded to a cast to `Int64`.


vermorel:
Sorry C# mistake, `long` does not mean the same thing is C++ as it does in C#.

PR #2082, opened 2017-07-05
'totalSamplesSeen' cast as Int64 instead of Int32 to avoid overflow

vermorel:
When performing long runtime training tasks, the
number `totalSamplesSeen` as printed by `cntk.exe`
does overflow at 2^31, and the console starts displaying
negative counts (situation actually observed). The
overflow is caused by a cast to `Int32` which can be
upgraded to a cast to `Int64`.


mahilleb-msft:
Thanks @vermorel  - could you perhaps change to %zu and drop the cast?

Thanks, Mark

vermorel:
@mahilleb-msft  Mark, thanks a lot. Indeed, it's much more idiomatic that way.

mahilleb-msft:
Thanks for updating it! I'll run this through our CI...

mahilleb-msft:
It's merged, thanks for your contribution!

PR #2090, opened 2017-07-06
Lock file path problem in CentOS distribution

hrasadi:
Because the /var/lock directory is not writable by users in CentOS, the binary releases of CNTK is practically unusable if you do not have root access on such machines. This proposal, relocates the lock file from /var/lock to /tmp that solves the permission problem completely and consistently among all linux distros.


Source/Common/CrossProcessMutex.h
```diff
@@ -128,7 +128,7 @@ class CrossProcessMutex
 public:
     CrossProcessMutex(const std::string& name)
         : m_fd(-1),
-          m_fileName("/var/lock/" + name)
+          m_fileName("/tmp/" + name)
```
wolfma61: > m_fileName("/tmp/" + name) [](start = 9, length = 27)

it is agreed that this issue needs a solution, but this should be solved in an user-configurable way, not through a static assignment

wolfma61:
:clock1:

PR #2103, opened 2017-07-10
Adding 'template' keyword (compiling with GCC)

imriss:
It is the 2nd instance in this file. The first instance was already fixed with the 'template' keyword.

Thanks.


zhouwangzw:
Thanks for your contribution. We are looking into the PR and will integrate it soon.

PR #2110, opened 2017-07-12
Porting to OpenPOWER clusters

doichanj:
I added __PPC64__ definitions to ignore SSE functions that is not available on POWER processors as similar to ARM processors.
(I think we have to add Altivec/VMX implementations, but currently I think these are not necessary if we have GPU)

I also found same problem while compiling Tests/UnitTests/MathTests/BlockMultiplierTests.cpp but I could not find the way to ignore SSE in this test program. I just commented out this test program compilation in the Makefile (But I did not commit this one)

### DIFF
--- a/Source/Common/Include/ssefloat4.h
+++ b/Source/Common/Include/ssefloat4.h
@@ -11,7 +11,7 @@
 #include <intrin.h> // for intrinsics
 #endif
 #ifdef __unix__
-#if !defined(__aarch64__)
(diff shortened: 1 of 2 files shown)


doichanj:
I also added Makefile for OpenPOWER

KeDengMS:
Hi @doichanj,

Thanks for the contribution. Could you take some time to sign the CLA?

Issue #2109, opened 2017-07-12
Implement hyperbolic functions and their inverses


lakshayg:
CNTK currently does not have the following functions

- [ ] `Sinh`
- [ ] `Cosh`
- [ ] `Asinh`
- [ ] `Acosh`
- [ ] `Atanh`

Some of them can be useful as activation functions (http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.10.6514&rep=rep1&type=pdf). I would like to implement this functionality in CNTK. Before starting with the implementation, I would like to have some feedback.

ishikasoni:
Can I work on implementing acosh

lakshayg:
@ishikasoni Sure! go ahead.

PR #2117, opened 2017-07-14
implement hyperbolic functions

lakshayg:
Refer #2109. Implements sinh and cosh.


bindings/python/cntk/ops/__init__.py
```diff
@@ -1454,6 +1454,50 @@ def cos(x, name=''):
...
 
+@typemap
+def sinh(x, name=''):
+    '''
+    Computes the element-wise sinh of ``x``:
+
+    The output tensor has the same shape as ``x``.
+
+    Example:
+        >>> np.round(np.sinh([[1,0.5],[-0.25,-0.75]]).eval(),5)
```
KeDengMS: Test failure. This should be np.round(C.sinh([[1,0.5],[-0.25,-0.75]]).eval(),5)

lakshayg: Thanks for the feedback! I just pushed a commit to fix this.

PR #2131, opened 2017-07-18
fix getitem for multioutput to fallback to outputs

boeddeker:


KeDengMS:
Thanks for the contribution! This is merged into master in https://github.com/Microsoft/CNTK/commit/7034de260ee20fc0923709c3c3371f8ff8958d45

PR #2154, opened 2017-07-24
Put gradient into a contiguous buffer for faster aggregate

FDecaYed:
This approach switches the underlying matrix of each gradient tensor with a slice of a contiguous buffer. So there won't be extra memcpy and all matrix should be put together.

This is done after the memory sharing is decided.
In best case, where these matrix are not shared with other matrices, the memory usage remain same.
In worst case, the memory consumption will be increased by some fraction of total size of gradient.

In cases there is not enough memory for contiguous buffer, it will fall back to not using it. So nothing will fail because of turning on this.

### DIFF
--- a/Source/Math/NcclComm.cpp
+++ b/Source/Math/NcclComm.cpp
@@ -22,12 +22,14 @@ static void operator||(cudaError_t rc, const char *msg)
 NcclComm::NcclComm(int deviceId, const MPIWrapperPtr& mpi)
     : m_ncclComm(nullptr), m_stream(nullptr)
 {
-    if (mpi->IsMultiHost())
-        return;
-
     size_t numRanks = mpi->NumNodesInUse();
     std::vector<int> allDevs(numRanks);
+    std::vector<std::array<char, MPI_MAX_PROCESSOR_NAME>> allHosts(numRanks);
+    std::array<char, MPI_MAX_PROCESSOR_NAME> procName {};
+    int nameLen;
+    MPI_Get_processor_name(procName.data(), &nameLen);
     mpi->Allgather(&deviceId, 1, MPI_INT, allDevs.data(), 1, MPI_INT);
+    mpi->Allgather(procName.data(), MPI_MAX_PROCESSOR_NAME, MPI_CHAR, allHosts[0].data(), MPI_MAX_PROCESSOR_NAME, MPI_CHAR);
 
     for (size_t r = 0; r<numRanks; r++)
     {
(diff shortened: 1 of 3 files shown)


junjieqian:
There are some packing logic in both v1 and v2 aggregator (https://github.com/Microsoft/CNTK/commit/a6e13ce96934354f9bae2fbe134071852e6340ee and https://github.com/Microsoft/CNTK/commit/bd93d7ec5a7a594d241e316e73be803a9868a904). Are these changes same as the commits?

FDecaYed:
While the goal is the same, these are two different approaches. There are several differences to my understanding.
1. There will be some extra memcpy in a6e13ce. Though it is possible to hide most of them, it is still possible to introduce speed overhead. Also, I remember there was a bug introduced by it related to cuda stream. I think it is fixed by now?
2. In a6e13ceh, there are two copy of matrices smaller than the threshold.(if I understand it right) It rely on user to set the right threshold to pack as much as matrices while not exceeding memory limit. The problem is, in order to pack all matrices, it requires at least the same amount of extra memory(potentially more) compare to the new proposed one.
3. In a6e13ce, though 'threshold' can be set, it is very hard for user to tune the threshold since they have no way of telling what's going on. The new approach is also not good in this point, but I think we can implement 'pack as much as possible'.

Overall, I think the problems is packing more matrices is better for speed but without a threshold it is more likely to fail. It should be fine for most network with smaller weights(convolution) but we may need to investigate more on both speed and memory trade off on other networks.

Also, it is mentioned in a6e13ce that change only apply when not using NCCL. I believe the change in this PR apply to both mpi and NCCL path.

junjieqian:
The arguments are correct that, 1. additional overhead introduced (but AFAIK the bug that data not ready on GPU is gone); 2. threshold value required for packing. Using a threshold is that we do not want pack large gradients, but as you said it needs tuning from user side. But without threshold, we may not be able to pack at all for some cases. With threshold, small gradients are packed except large ones. In some of our benchmarks, for example, 5 out of 7 gradients can be packed.

The drawback of current packing logic is that there are too many conditions. According to my experiment results, packing-all (without threshold) is much faster than current approach. So we can improve this packing logic further. BTW, do you have numbers that with your packing, how much is the performance improvement?

We could enable packing for NCCL as well, which is straightforward, but at that time the concern is that NCCL is already fast enough, do we want to introduce packing overhead in it?

FDecaYed:
I think now we are down to 2 problem.
1. Which matrices are being packed?
 - You have a pack matrices less than threshold method.
 - This PR uses 'pack everything if possible, otherwise pack nothing'. We choose this approach only because it is the easiest to do.
 - I think 'pack as much as possible after sort' may be the best performing solution. But given that memory usage in other part can be dynamic, it is tricky and potentially hard.
2. How to pack
I think the approach used here by change pointer is better on both speed and memory than the memcpy approach. Though it looks hacky without changing other parts of cntk code.

PR #2186, opened 2017-08-03
Tweaks for cuda9 build support

FDecaYed:
Contains following changes:
- build support for cub1.7, note that build with 1.4 will break
  cuda 9 introduced new threading model, which will break older cub. updating to 1.7 allow build with both cuda 9 and earlier.
- support for multi-node initialization in NCCL2
- check for cudnn version to allow build with both 6&7 because of API deprecation

### DIFF
--- a/Source/Math/CntkBatchNormalization.cuh
+++ b/Source/Math/CntkBatchNormalization.cuh
@@ -107,12 +107,16 @@ __device__ __forceinline__ void StoreValues<4, float>(const float src[4], float*
 }
 
 template <typename T>
-__device__ __forceinline__ T Shuffle(T input, int srcLane)
+__device__ __forceinline__ T Shuffle(T input, int srcLane, unsigned int mask)
 {
 #ifdef __CUDA_ARCH__
     // shfl is supported only on Kepler+
     static_assert(__CUDA_ARCH__ >= 300, "CNTK only supports only Kepler GPU architecture or newer.");
+#if CUDA_VERSION >= 9000
+    return cub::ShuffleIndex(input, srcLane, CUB_PTX_WARP_THREADS, mask); // Need cub > 1.7.0
+#else
     return cub::ShuffleIndex(input, srcLane);
+#endif
 #else
     assert(false);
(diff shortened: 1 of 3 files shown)


mahilleb-msft:
Hi @FDecaYed, our Windows build doesn't like the #if / #endif around just the "CUDNN_CALL(.." line. Seems we need to duplicate the whole CUDNN call parameters here. Also, CUB 1.7.0 generates lots of 4706 warnings on Windows, one way to suppress them is by changing the compile options in MathCUDA.vcxproj to ignore it.
```
<AdditionalOptions>-Xcudafe "--diag_suppress=field_without_dll_interface" -Xcompiler "/wd 4819 /wd 4706" %(AdditionalOptions)</AdditionalOptions>
```
I'll check if there are alternatives.

Would there a way to preserve backwards compatibility to the old CUB library for CUDA < 9.0? If this has to be a breaking update, we'll need to careful on timing when integrating this.

Thanks, Mark

FDecaYed:
@mahilleb-msft Hi mark,
there is no way that I know of to check cub version. I updated the PR to do the following:
```
#if cuda9
call 1.7 api
#else
call old cub api
```
This way, things remain the same in cuda8. And user can only compile with cub1.7 if the are using cuda9. Then this PR is no longer a breaking update and can be integrated.

On the other hand, changing default to cub1.7 and only support one version of cub is still worth doing and I agree we need to test and find timing on that.

Attach link to libraries per @cha-zhang 's request.
Cuda 9.0RC – https://developer.nvidia.com/cuda-release-candidate-download
NCCL 2.0.4 – https://developer.nvidia.com/nccl
CUB 1.7 – https://github.com/NVlabs/cub/archive/1.7.0.zip

FDecaYed:
Just rebased this to TOT.
Fixed a warp synchronize issue that causes hang on volta(shuffle in CntkBatchNormEngine).
Also squashed all changes into one commit.

KeDengMS:
Thanks @FDecaYed! This change has been merged in master and available in 2.4 release.

PR #2280, opened 2017-08-29
numeric_limits::min returns positive value for floats - breaks argmax

DaveyBiggers:
This fixes a bug with the CPU implementation of argmax.
To reproduce original bug (in Python):

```
import cntk
data = [-4,-5,-6]
cntk.argmax(data).eval()
```
expected: 0
actual: -1

### DIFF
--- a/Source/Math/CPUMatrixImpl.h
+++ b/Source/Math/CPUMatrixImpl.h
@@ -7332,12 +7332,12 @@ template <class ElemType>
 int CPUMatrix<ElemType>::Argmax() const
 {
     int maxArg = -1;
-    ElemType maxValue = std::numeric_limits<ElemType>::min();
+    ElemType maxValue = std::numeric_limits<ElemType>::lowest();
 
 #pragma omp parallel 
     {
         int localMaxArg = -1;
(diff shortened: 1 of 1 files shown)


Source/Math/CPUMatrixImpl.h
```diff
@@ -7332,12 +7332,12 @@ template <class ElemType>
 int CPUMatrix<ElemType>::Argmax() const
 {
     int maxArg = -1;
-    ElemType maxValue = std::numeric_limits<ElemType>::min();
+    ElemType maxValue = std::numeric_limits<ElemType>::lowest();
```
ebarsoumMS: Thank you for the fix, can you update ArgMax python unit test to cover the above case?

The unit test is in: https://github.com/Microsoft/CNTK/blob/master/bindings/python/cntk/ops/tests/reduction_test.py

DaveyBiggers: Apologies, I don't have the bandwidth to get set up with a CNTK dev environment - I could modify the unit test, but wouldn't be able to run it.

ebarsoum: No problem I will take care of it.

PR #2429, opened 2017-09-29
fix building issus with cub 1.7.4(or 1.7.x) and cudnn 7

LinkeyLeo:


bencherian:
Just FYI this appears to be very similar to #2186. I imagine this won't get merged until CUDA 9 support is added.

LinkeyLeo:
while this issue also occurs when using cuda 8, it not caused by cuda while cudnn and cub

bencherian:
I'm aware that this problem is not directly caused by CUDA 9, but there's no particular need to use CUB 1.7 without CUDA 9 and changing the function call will break compatibility with older versions (and there's no good way to do compiler detection of CUB versions)...I think the proposal in the other PR of tying CUB 1.7 use to CUDA 9 is a reasonable solution for this.

LinkeyLeo:
there are two different issue, one is about a func calling in cub1.7, the other is about cudnn 7

PR #2500, opened 2017-10-16
Modified ScaleTransForm

calledbymountains:
Added Options "ResizeMin" and "ResizeMax" to ScaleTransform.

ResizeMin  sets the smaller dimension of the image to a fixed size while keeping the same aspect ratio.
ResizeMax sets the larger dimension of the image to a fixed size while keeping the same aspect ratio.

This allows us to use a greater diversity during data augmentation.

This method has been used in the [OverFeat Paper](https://arxiv.org/abs/1312.6229)


calledbymountains:
The commit has been tested successfully.

PR #2562, opened 2017-10-28
Changed setting in selecting VGGNet16 as base_model for Fast/Faster R-CNN

kyoro1:
We cannot obtain Fast/Faster R-CNN with current parameter settings under VGGNet16 base model.
(Indeed, loss of cross-entropy will be "na", because they're too big values.)

These changes include modifications, which enable us to get correct results under VGGNet16.


Examples/Image/Detection/utils/configs/VGG16_config.py
```diff
@@ -23,7 +23,9 @@
 __C.MODEL.FEATURE_STRIDE = 16
 __C.MODEL.RPN_NUM_CHANNELS = 512
 __C.MODEL.ROI_DIM = 7
-__C.MODEL.E2E_LR_FACTOR = 1.0
-__C.MODEL.RPN_LR_FACTOR = 1.0
+## In using Faster R-CNN, try E2E_LR_FACTOR or RPN_LR_FACTOR from 0.1 or smaller.
+## They decide learning rate in DNN.
+__C.MODEL.E2E_LR_FACTOR = 0.1
```
spandantiwari: Which dataset are you setting these numbers for? Grocery, Pascal VOC 2007? These settings are directly taken from FasterRCNN paper, which works with VOC2007, 2012 data. CNTK's goal is to showcase the training as matching that in the paper which is why I don't think these parameters should be changed in the master. You can change them in your own runs.

Examples/Image/Detection/utils/configs/VGG16_config.py
```diff
@@ -23,7 +23,9 @@
 __C.MODEL.FEATURE_STRIDE = 16
 __C.MODEL.RPN_NUM_CHANNELS = 512
 __C.MODEL.ROI_DIM = 7
-__C.MODEL.E2E_LR_FACTOR = 1.0
-__C.MODEL.RPN_LR_FACTOR = 1.0
+## In using Faster R-CNN, try E2E_LR_FACTOR or RPN_LR_FACTOR from 0.1 or smaller.
+## They decide learning rate in DNN.
```
spandantiwari: "They decide learning rate in DNN." - Is this comment for the user? If it is not absolutely necessary, please remove the comment.

Examples/Image/Detection/utils/configs/VGG16_config.py
```diff
@@ -23,7 +23,9 @@
...
+__C.MODEL.E2E_LR_FACTOR = 0.1
+__C.MODEL.RPN_LR_FACTOR = 0.1
```
spandantiwari: This too.

Examples/Image/Detection/utils/configs/VGG16_config.py
kyoro1: I see. But, as far as I tried with Grocery data without parameter setting, `loss` was `nan` which means loss is very huge... So, it'd be better to change to converge even with default parameters, I assumed.

spandantiwari: Hmm, interesting that you are seeing NaN with grocery dataset. In any case, it is assumed that this would be changed for different datasets. But setting it as a default based on just one dataset (in this case, grocery) wouldn't be wise. If there's a better default, it must be tested on multiple sets, including VOC 2007.

spandantiwari: Yes, a comment here is the right way to go. Something broad like "Try changing this parameter if the training does not converge. E.g. for Grocery dataset, it may be better to set this to 0.1."

kyoro1: Thanks. So, I'll change the codes, and it could be helpful. Indeed, when I tried Grocery example and loss was `nan`, I was at a loss...

kyoro1:
I'll update with only adding the comments. Close the issue.

spandantiwari:
OK, sounds good.

PR #2627, opened 2017-11-17
Fix libzip download url

mogmod:
http://nih.at/libzip is redirected to https://libzip.org


KeDengMS:
Thanks for the contribution!

PR #2628, opened 2017-11-17
Fix Dockerfiles --with-mkl path for Intel MKL-DNN(2017-11-10).

mogmod:
Fix Dockerfiles --with-mkl path for Intel MKL-DNN(2017-11-10).


KeDengMS:
Thanks for the contribution!

Issue #2388, opened 2017-09-21
”CUDA failure 77: an illegal memory access was encountered" in Executing Faster R-CNN


kyoro1:
In executing `run_faster_rcnn.py` of CNTK2.2, I encountered error message as follows. With version=2.1, I've never met this kind of error.

- Environment:
  - NC24 in Azure(4-GPU)
  - CNTK: 2.2
  - python 3.5

What kind of activities can we do in this situation? (As far as I tried, Fast R-CNN scripts was fine in the same environment.)

```sh
C:\git\CNTK\Examples\Image\Detection\FasterRCNN (master)
(cntk2.2) λ python run_faster_rcnn.py
Selected GPU[1] Tesla K80 as the process wide default device.
Using base model:   AlexNet
lr_per_sample:      [0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.001, 0.0001, 0.0001, 0.0001, 0.0001, 0.0001, 0.0001, 0.0001, 0.0001, 0.0001, 0.0001, 1e-05]
Training model for 20 epochs.
Training 57554122 parameters in 27 parameter tensors.
-------------------------------------------------------------------
Build info:

                Built time: Sep 15 2017 07:42:54
                Last modified date: Thu Sep 14 22:33:54 2017
                Build type: Release
                Build target: GPU
                With 1bit-SGD: no
                With ASGD: yes
                Math lib: mkl
                CUDA version: 8.0.0
                CUDNN version: 6.0.21
                Build Branch: HEAD
                Build SHA1: 23878e5d1f73180d6564b6f907b14fe5f53513bb
                MPI distribution: Microsoft MPI
                MPI version: 7.0.12437.6
-------------------------------------------------------------------
CUDA failure 77: an illegal memory access was encountered ; GPU=0 ; hostname=ABC-XYZ ; expr=cudaMemcpy(Data(), pArray, sizeof(ElemType) * GetNumElements(), (matrixFlags & matrixFlagSetValueOnDevice) ? cudaMemcpyDeviceToDevice : cudaMemcpyHostToDevice)
Traceback (most recent call last):
  File "run_faster_rcnn.py", line 35, in <module>
    trained_model = train_faster_rcnn(cfg)
  File "C:\git\CNTK\Examples\Image\Detection\FasterRCNN\FasterRCNN_train.py", line 291, in train_faster_rcnn
    eval_model = train_faster_rcnn_e2e(cfg)
[...]

PR #2718, opened 2017-12-03
distributed Faster R-CNN modules

kyoro1:
Hi @KeDengMS @spandantiwari @ebarsoumMS , I tentatively generated distributed Faster R-CNN modules with "distributed" module in cntk.
Also, in these module, I use cpu_nms with the comment from @KeDengMS in [the issue](https://github.com/Microsoft/CNTK/issues/2388#event-1368178014). This is because I understood that we would use cpu_nms as default with Faster R-CNN so far.

Anyway, I'll squash these commits after your verification as usual, and please review these scripts.


kyoro1:
Just in case, with these scripts, I have succeeded in the following cases:

- Environment: 
  - cntk version: 2.3
  - python: 3.5
  - Azure: Windows NC12 with 2 GPU
  - path: \CNTK\Examples\Image\Detection\FasterRCNN
  - dataset: Grocery

1) normal case
`python run_faster_rcnn.py` 

2) distributed learning
`mpiexec -n 2 python run_faster_rcnn.py`

PR #2720, opened 2017-12-03
fix readonly bug in Value::Create(); adapt and add tests

zhouwangzw:
Bug description: When creating a Value from the provided vector of NDArrayView while readonly is set to false, the returned Value object does not have the expected readonly property. Instead of a non-readonly as the parameter requires, the returned Value object is always read-only. 

This PR fixes the bug and also adds new tests for checking the readonly property. 
The PR is based on release/2.2. Please rebase to the current master before merging.


zhouwangzw:
Close this one, as I have created another one which can be merged directly into master.

PR #2754, opened 2017-12-10
Changes required to build on Ubuntu 16.04/gcc 5.4, plus javadoc

pgarst:
These are a few small changes required to build the current CNTK system on Ubuntu 16.04/gcc 5.4.

Trainer.cpp required one cast to satisfy the 5.4 compiler.

With gcc the current code fails with c++11 with this error: lambda capture initializers only available with -std=c++14.
Also, cuda 8 does not support c++14. To make it work I had to use c++14 for gcc and c++11 for nvcc.

I also added a line to the Makefile to generate javadoc documentation for the Java wrapper.


sk-g:
this solves #2800

Makefile
```diff
@@ -115,7 +115,7 @@ SRC:=
 all : buildall
 
 # Set up basic nvcc options and add CUDA targets from above
-CUFLAGS = -m 64
+CUFLAGS = -m 64 -std=c++11
```
sk-g: This works. But is it because NVCC does not support c++14 (?)

pgarst: According to the docs, Cuda 9 supports c++14 but Cuda 8 only supports c++11. I believe CNTK still requires Cuda 8, not 9, but when that changes we could make the standard uniform again.

KeDengMS:
Can you merge with master and check if this fix is still needed?

pgarst:
I’m currently traveling. When I get back about 2/21 I’ll pull the master
and see how much of this is still needed, unless someone else has resolved
it before then.

Peter

KeDengMS:
CNTK 2.4 release has switched to Ubuntu 16.04 / gcc 5.4.

PR #2759, opened 2017-12-11
Bug fixes: fp16 default behavior

FDecaYed:
This line of commit message is not completely accurate:
set convolution mathType to tensor core only for fp16, and normal math for fp32

The change is actually trying to address this problem:
cudnn find will give you the fastest algo by number and its math type, but before find, we need to set allow tensor core math in order to include algo using tensor core in our searching.
After find, say algo1 without tensor core math is the fastest, we need to record it correctly and set mathtype to disable tensor core, otherwise we will be running algo1 with tensor core.

There might be bug in those batchsize changing cases that cntk allow, but all these are within if(half), so fp32/64 should not be affected.


KeDengMS:
Thanks @FDecaYed , merged

PR #2805, opened 2017-12-26
Fixes #2704

### LINKED ISSUE #2704: Makefile needs some fix
Hi,

`#` should be removed at [line 691](https://github.com/Microsoft/CNTK/blob/master/Makefile#L691), [line 696](https://github.com/Microsoft/CNTK/blob/master/Makefile#L696), and [line 700](https://github.com/Microsoft/CNTK/blob/master/Makefile#L700). Otherwise, CPU only version will be used regardless of `CUDA_PATH` being defined or not.

Thanks.

nglee:
Without this fix, CPU only version will be used regardless of CUDA_PATH being defined or not.


manikjindal:
Thanks @nglee for making this fix.

Issue #2784, opened 2017-12-18
Add support for explicit weight decay


bencherian:
There was recently a [paper posted on arxiv ](https://arxiv.org/abs/1711.05101) which gives some plausible explanations for why adaptive gradient methods like Adam tend to not generalize as well as SGD or SGD+momentum. This relates to how L2 regularization affects weight updates. The authors propose explicitly adding a weight decay to a standard adaptive gradient update instead of including it as a L2 regularization term. Would it be possible to add an option to do something like this for the adaptive gradient learners in CNTK? (I might be able to get some time to implement this in a PR. If you'd be interested in accepting this, would it make more sense to have a separate parameter for explicit weight decay or a flag that turns L2 regularization loss into a pseudo-equivalent weight decay?)

bencherian:
I added support for this in the referenced PR #2810. The current implementation is kind of ugly because it sticks the weight decay value(s) on the end of the parameter list for all the learners in the Python API, but I don't really see a way around it. Also, I'm not entirely sure how you would add similar support to other APIs (I'm not familiar enough with C# or Java to really figure out what to change). Are there any thoughts on how this could be implemented?

PR #2810, opened 2017-12-30
Add explicit weight decay

bencherian:
Adds explicit support for weight decay as per request in #2784.
This enables better optimization behavior for adaptive gradient methods (e.g., adagrad, adam) and easier hyperparameter tuning all techniques outside of vanilla SGD due to improved separation between learning rate and regularization terms. In order to enable weight decay schemes discussed in the [paper](https://arxiv.org/abs/1711.05101) which highlights this issue, the user is able to specify the weight decay using a schedule.


bencherian:
Sorry, I just realized the code pushed to this branch is broken...I'll update this shortly.

bencherian:
Ok, this should actually run and pass tests now.

bencherian:
Would it be possible to get some feedback on whether this PR might be merged?

PR #2840, opened 2018-01-15
Small changes for CUDA 9 compatibility

BenjaminTrapani:
Update VS build to use CUDA 9 if present. Update CuDnnRnn call to match updated version in CUDA 9. These changes simply support CUDA 9 but do not leverage any of the new functionality besides performance improvements to previously leveraged APIs.


BenjaminTrapani:
It looks like https://github.com/Microsoft/CNTK/pull/2186 has covered these changes in a backwards-compatible way, will use those changes.

PR #2880, opened 2018-01-25
Fix TensorBoard on_write_test_summary method

### LINKED ISSUE #2878: TensorBoard test summary error
Bug report, version CNTK 2.3.1

The _TensorBoardProgressWriter_ python class is causing a fatal error in the method _on_write_test_summary_ line [530](https://github.com/Microsoft/CNTK/blob/1cca6e2c1742fa399e8ec2bab5604a2808530f95/bindings/python/cntk/logging/progress_print.py#L522-L530).

Error caused:
```
AttributeError: 'ProgressWriter' object has no attribute 'summaries'
``` 

Observations:
It seems that where it says "self.summaries" should say "summaries" which is one of the method's arguments. I tried this fix on my side and it seems to work as expected. 

Let me know if it's OK to send a PR.

rnrneverdies:
Fixes #2878, TensorBoardProgressWriter causes a fatal error while trying to write the summary test average (at the end of the training session).

### DIFF
--- a/bindings/python/cntk/logging/progress_print.py
+++ b/bindings/python/cntk/logging/progress_print.py
@@ -527,7 +527,7 @@ class TensorBoardProgressWriter(cntk_py.ProgressWriter):
             # This allows to easier correlate the training and test metric graphs in TensorBoard.
             self.write_value('minibatch/test_avg_metric', avg_metric, self.total_training_updates())
         else:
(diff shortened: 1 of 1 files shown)


akanimax:
This doesn't seem to be merged yet. I am facing the same problem :(

PR #2924, opened 2018-02-05
Fix BinaryConvolutionLib project

igormq:
* Fix project dependency
* Add Halide.dll post build script
* Modify vsbuild.bat
* Remove unnecessary spaces


bencherian:
There are HalidLibPath and HalideIncludePath macros defined in CNTK.cpp.props. Have you considered using those instead? I would also think it makes sense to link to the debug Halide library when making a debug CNTK build.

igormq:
Hi @bencherian, thank you for your comments.

Yes, indeed there is HalideLibPath, however, the path only points to Release DLL  (e.g., C:\local\Halide\Release) and the HalideIncludePath points to the folder that has the Halide.h file. 

I linked the debug/release Halide library making the debug/release build

PR #3021, opened 2018-03-08
MKLDNN primitive support for VGG model

zhenlinluo:
Ke, pls review all the patches for MKL DNN primitives integration including conv, batchnorm, pooling, relu, timestranspose. To get VGG model fully accelerated, we also add convolution engine features for bias and relu. Pls review all of them and provide your feedback. Thanks


KeDengMS:
This is currently handled in branch kedeng/mkldnn. Will verify with tests and merge from there to master.

PR #3029, opened 2018-03-10
Image transformer fixes

BenjaminTrapani:
If an image of size 50 x 100 was scaled using the pad mode to a size of 20 x 100, the resulting image would be incorrect, since it would be scaled along the Y axis when it should have been scaled along the X axis. This was a result of the logic that selected the scaled dimension always selecting the larger dimension of the source image. This works if the source image is smaller than or equal to the destination size, but does not always work correctly when shrinking to a target size.

OpenCV 3.4 support is added because this version fixes an issue where exif parsing errors would cause the application to hang when decoding a jpeg file.


Source/Readers/ImageReader/ImageTransformers.cpp
```diff
@@ -368,54 +370,56 @@ StreamInformation ScaleTransformer::Transform(const StreamInformation& inputStre
...
         else
-            scaleW = width > height; // else we resize the larger side
+            scaleW = widthFactor < heightFactor; // else we resize along the smaller ratio
 
         size_t targetW, targetH;
         if (scaleW)
         {
-            targetW = (size_t)m_imgWidth;
-            targetH = (size_t)round(height * m_imgWidth / (double)width);
+            targetW = (size_t) m_imgWidth;
```
spandantiwari: Can you please remove changes like this one (and several others below) where you are introducing space in the casting operation. The original code is fine as far as coding style goes, and it is sort of obscuring the real changes in this PR.

CNTK.Cpp.props
```diff
@@ -9,7 +9,7 @@
     <NvmlDll Condition="Exists('c:\local\nvsmi9\NVSMI\nvml.dll')">c:\local\nvsmi9\NVSMI\nvml.dll</NvmlDll>
 
     <HasOpenCv>false</HasOpenCv>
-    <HasOpenCv Condition="Exists('$(OPENCV_PATH)') Or Exists('$(OPENCV_PATH_V31)')">true</HasOpenCv>
+    <HasOpenCv Condition="Exists('$(OPENCV_PATH)') Or Exists('$(OPENCV_PATH_V31)') Or Exists('$(OPENCV_PATH_V34)')">true</HasOpenCv>
```
spandantiwari: Is this really needed in this PR? If not, I think we should undo this and leave only one version of OpenCV in there (the one that we have listed on our installation page).

BenjaminTrapani:
@spandantiwari How do the changes look?

PR #3124, opened 2018-04-15
change cprintf to use vsnprintf

libohanr100:
In linux, it will write file, which will cause slow perf, how about change to this?

### DIFF
--- a/Source/Common/Include/Basics.h
+++ b/Source/Common/Include/Basics.h
@@ -230,14 +230,17 @@ private:
         return _vscwprintf(format, args);
 #elif defined(__UNIX__)
         // TODO: Really??? Write to file in order to know the length of a string?
-        FILE* dummyf = fopen("/dev/null", "w");
-        if (dummyf == NULL)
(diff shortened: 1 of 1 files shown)


Source/Common/Include/Basics.h
```diff
@@ -230,30 +230,30 @@ struct _strprintf : public std::basic_string<_T>
...
-        FILE* dummyf = fopen("/dev/null", "w");
-        if (dummyf == NULL)
-            perror("The following error occurred in basetypes.h:cprintf");
-        int n = vfwprintf(dummyf, format, args);
-        if (n < 0)
-            perror("The following error occurred in basetypes.h:cprintf");
-        fclose(dummyf);
-        return n;
+		const int BUF_SIZE = 256;
+		int n = 0;
```
KeDengMS: Hard tabs should be replaced with spaces. Currently the merge is blocked in CI because of this.

libohanr100:
why linux build failed?

KeDengMS:
Seems a issue in our build system to pick up the commit. @thiagocrepaldi can you take a look?

PR #3192, opened 2018-05-09
Handle intermediate directory creation properly

AnmolMago:
This simple change allows python to create intermediate directions if they do not exist.


kit1980:
I don't see `os.mkdirs()` in Python. You probably meant `os.makedirs()`?

PR #3206, opened 2018-05-17
fix break for cub 1.8 API change

FDecaYed:
Note that we don't support multiple cub version. This code keep this behavior and only support cub version >= 1.8.0


KeDengMS:
Seems we need some build side changes to take this. @thiagocrepaldi please take a look when you got time.

PR #3245, opened 2018-06-06
Make it to build in VS.

rpengms:


rpengms:
Make it to build in VS.

PR #3290, opened 2018-07-07
add in new sequence layer: QRNN

delzac:
Added in a new sequence layer, Quasi-Recurrent Neural Network.

### DIFF
--- a/bindings/python/cntk/layers/sequence.py
+++ b/bindings/python/cntk/layers/sequence.py
@@ -12,6 +12,7 @@ from ..variables import Record
 from ..ops import combine, splice, sequence, reconcile_dynamic_axes
 from .blocks import *
 from .blocks import _get_initial_state_or_default, _inject_name
+from .layers import SequentialConvolution
 
 
 def Delay(T=1, initial_state=default_override_or(0), name=''):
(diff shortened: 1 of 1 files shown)


bindings/python/cntk/layers/sequence.py
```diff
@@ -666,3 +666,65 @@ def unfold_from(initial_state, dynamic_axes_like):
...
+    """
+
+    def FoPool(c, fz):
+        f = C.slice(fz, 0, 0, hidden_dim)
+        z = C.slice(fz, 0, hidden_dim, 2 * hidden_dim)
+        return f * c + (1 - f) * z
+
+    def model(input_tensor):
+        gate_values = Convolution(filter_shape=filter_shape, num_filters=3 * hidden_dim, pad=True,
+                                  reduction_rank=1, sequential=True)(input_tensor)
```
BowenBao: Recently we have updated and added a new interface C.layer.SequentialConvolution([link](https://github.com/Microsoft/CNTK/blob/3806ec6e0ebb3f47ac56cd937a5b9f3dcf3108d8/bindings/python/cntk/layers/layers.py#L278)) for more complete support of sequential convolution. It'd be great if you could have a look and switch to the new api. It should work the same by switching the called method name to SequentialConvolution and drop the parameter "sequential=True".

delzac: Updated to the new API. Thanks for pointing that out!

Issue #3043, opened 2018-03-17
RuntimeError: ONNX (Upsample) is not supported in CNTK


alexionby:
I tried to run PyTorch U-net model in CNTK through ONNX, but got this message.
And i got the same message with TransposedConv.

How can i avoid this?
```

Traceback (most recent call last):
  File "onnx_test.py", line 2, in <module>
    z = C.Function.load("our_unet.onnx", device=C.device.cpu(), format=C.ModelFormat.ONNX)
  File "/home/alex/anaconda3/lib/python3.6/site-packages/cntk/internal/swig_helper.py", line 69, in wrapper
    result = f(*args, **kwds)
  File "/home/alex/anaconda3/lib/python3.6/site-packages/cntk/ops/functions.py", line 1608, in load
    return cntk_py.Function.load(str(model), device, format.value)
RuntimeError: ONNX (Upsample) is not supported in CNTK

[CALL STACK]
[0x7fc854975ef9]                                                       + 0x8aeef9
[0x7fc854bfc05d]    CNTK::ONNXToCNTKHelper::  CreateFunction  (ONNXIR::Node const*,  std::vector<CNTK::Variable,std::allocator<CNTK::Variable>> const&) + 0x874d
[0x7fc854bfd48e]    CNTK::ONNXToCNTKHelper::  CreateCNTKNode  (ONNXIR::Node const*,  std::vector<CNTK::Variable,std::allocator<CNTK::Variable>> const&,  CNTK::DeviceDescriptor const&) + 0x1ce
[...]

PR #3335, opened 2018-08-02
Fix ONNX import issue with ConvTranspose operation without output_shape

cbecker:
This fixes a ONNX import issue that was mentioned in #3043 

The current master branch throws a NOT_IMPLEMENTED exception if the ONNX node doesn't have a `output_shape` attribute, while the CNTK ConvTranspose operator can be invoked without such knowledge.

This PR allows for a ONNX node that can optionally contain the `output_shape` attribute.

NOTE: I've also refactored the code according to what I interpreted that is wanted here.


cbecker:
Is there interest in merging this PR?

spandantiwari:
@cbecker Thanks for creating the PR. FYI there's a big overhaul to ConvTranpose coming in. It has been in the works for a few days and it should be in master soon (a week). 
In addition to output_shape, it also supports pads, output_padding, auto_pad attributes. 
If you could test it out once its there and let us know of any issues, that'll be much appreciated. 

Thanks!

PR #3362, opened 2018-08-14
Adds convolution registry to cache cudnn find convolution calls

fran6co:
Cuts down warmup time by a lot, in my use case it went from 26 seconds to 6. And it avoids having to re autotune when running the network in parallel


Source/Math/CuDnnConvolutionEngine.cu
```diff
@@ -206,6 +206,198 @@ enum class AutotuningState : int
...
+
+    size_t DeterministicAlgoWorkspaceSize; // workspace size for deterministic algorithm
+
+    AutotuningState autotuningState; // state of auto-tuning: Init, PendingTuning and Running
+    int32_t selectedAlgo;            // currently selected algorithm
+    int32_t maxAlgo;                 // algorithm that was selected when the current workspace is allocated
+
+    cudnnMathType_t AlgoMathType;
+
+    bool NeedAutotuning(size_t batchSize, size_t workspaceSize)
```
WillBrennan: `bool NeedAutotuning(size_t batchSize, size_t workspaceSize) const;`

PR #3396, opened 2018-09-13
Fmeasure

delzac:
added a small constant to ensures correct results when target is empty i.e. all pixel are "background"


delzac:
@KeDengMS For you attention to include this PR before 2.6.0 is released.

PR #3414, opened 2018-09-26
fix network graph plotting for Python > 3.4

hjl:
The network graph plotting in CNTK requires pydot_ng, which is unmaintained after Python 3.4. Pydotplus is a drop in replacement, which enables network diagrams when using Python 3.5+.


prmtl:
I just wanted to let you know that pydot_ng is working on Python versions above 3.4 (there are CI jobs for 2.7, 3.4, 3.5, 3.6, 3.7 and even development version of 3.8)

hjl:
OK, I came across this while writing up Anaconda-based install instructions for some tutorials. There isn't a current conda package published for pydot_ng, but it can be installed by pip instead. CNTK itself needs to be installed using pip, so I guess one more isn't too messy.

PR #3530, opened 2018-12-11
Add raise statement before ValueError

daniel-s-ingram:


bindings/python/cntk/ops/functions.py
```diff
@@ -1466,7 +1466,7 @@ def train(self, minibatch_source,
                     else:
                         config[0] = cb
             else:
-                ValueError('callbacks list can only contain objects of type ProgressWriter, CheckpointConfig, CrossValidationConfig, and TestConfig.')
+                raise ValueError('callbacks list can only contain objects of type ProgressWriter, CheckpointConfig, CrossValidationConfig, and TestConfig.')
```
shahzadlone: Is there a benefit?

PR #3560, opened 2019-01-25
Guoye/merge embr final

guoli-ye:
Add the EMBR code for sequence training: EMBR is WER based sequence training method, it has been used for all MS ASR production model with very good gains.

### DIFF
--- a/Source/Common/Include/latticearchive.h
+++ b/Source/Common/Include/latticearchive.h
@@ -67,7 +67,7 @@ enum mbrclassdefinition // used to identify definition of class in minimum bayes
 // ===========================================================================
 class lattice
 {
-
+public:
     // definie structure for nbest EMBR
     struct TokenInfo
     {
--- a/Source/Common/Include/latticearchive.h
+++ b/Source/Common/Include/latticearchive.h
@@ -23,7 +23,7 @@
 #include <algorithm> // for find()
 #include "simplesenonehmm.h"
 #include "Matrix.h"
-
+#include <set>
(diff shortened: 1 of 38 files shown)


Source/Common/Include/latticearchive.h
```diff
@@ -23,7 +23,7 @@
 #include <algorithm> // for find()
 #include "simplesenonehmm.h"
 #include "Matrix.h"
-
+#include <set>
```
vmazalov: Do we need them sorted? Otherwise we can use unordered_set

Source/Common/Include/latticearchive.h
```diff
@@ -67,7 +67,28 @@ enum mbrclassdefinition // used to identify definition of class in minimum bayes
...
+        size_t prev_edge_index;
+        size_t prev_token_index;
+        double path_score; // use pure to indicatethe path score does not consider the WER of the path
+    };
+
+    struct NBestToken
+    {
+        // for sorting purpose
+        // make sure the map is stored with keys in descending order
+        std::map<double, std::vector<PrevTokenInfo>, std::greater <double>> mp_score_token_infos; // for sorting the tokens in map
```
vmazalov: why not ordered_map? -- similarly in other places.

guoli-ye: the map is for ordering purpose

Source/Common/Include/latticearchive.h
guoli-ye: change to unordered set

PR #3573, opened 2019-02-13
Rpengms/fp16 bmuf

rpengms:
Fix FP16 BMUF bug for V2/python.
Local learner's smoothed gradients now is a compound metrix of fp32 SG+TempGradients+Parameters. For BMUF:
1. ResetSmoothedGradients should NOT reset Parameters.
2. fp32 Parameters needs to be updated after BMUF distributed learner updated the fp16 ones.

### DIFF
--- a/Source/1BitSGD/BlockMomentumDistributedLearner.h
+++ b/Source/1BitSGD/BlockMomentumDistributedLearner.h
@@ -524,38 +524,40 @@ namespace CNTK
                     ResetBuffer<double>(i, p);
                 else if (p->GetDataType() == DataType::Float)
                     ResetBuffer<float>(i, p);
+                else if (p->GetDataType() == DataType::Float16)
+                    ResetBuffer<half, float16>(i, p);
                 else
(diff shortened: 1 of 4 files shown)


rpengms:
Those builds failed due to not enough disk space on build machines.

Source/1BitSGD/BlockMomentumDistributedLearner.h
```diff
@@ -470,6 +470,11 @@ namespace CNTK
 
             m_numSamplesSeenInCurrentBlock = 0;
 
+            // For half, SynchronizeModel will update the parameters (half) in network.
+            // However, the local learner also have a copy of full precision parameters (float), which needs to be updated too.
+            // Set the flag so that the float copy will be updated / copied from half parameters.
+            m_learner->SetNeedToUpdateMasterParameter();
```
KeDengMS: Should this be in the if (... == DataType:Float16) above, right after SynchronizeModel? float/double don't need it.

Issue #2913, opened 2018-02-02
Initialize weights - easy example


FlyinTeller:
When I try to initialize the weights and bias parameters for a Dense layer using the Python API I am getting a TypeError:

    l = c.layers.Dense(1000, activation=c.sigmoid, init=weights, init_bias=bias)(input_var)
      File "C:\Users\320000451\AppData\Local\Continuum\miniconda3\envs\py36\lib\site-packages\cntk\layers\layers.py", line 138, in Dense
    init_weights = _initializer_for(init, Record(output_rank=output_rank))
    File "C:\Users\320000451\AppData\Local\Continuum\miniconda3\envs\py36\lib\site-packages\cntk\layers\blocks.py", line 51, in _initializer_for
    init = initializer_with_rank(init, **rank_params)
    File "C:\Users\320000451\AppData\Local\Continuum\miniconda3\envs\py36\lib\site-packages\cntk\initializer.py", line 184, in initializer_with_rank
    return cntk_py.random_initializer_with_rank(initializer, output_rank, filter_rank)
    TypeError: in method 'random_initializer_with_rank', argument 1 of type 'CNTK::ParameterInitializer const &'

Here is what I am doing:

    import cntk as c
    input_var = c.input_var((180,))
    l = c.layers.Dense(1000, activation=c.sigmoid, init=c.glorot_uniform())(input_var) #create Dense layer
    weights = l.W.value # extract weights from layer
    bias = l.b.value        # extract bias from layer
    k = c.layers.Dense(1000, activation=c.sigmoid, init=weights, init_bias=bias)(input_var) #try to create new layer with same weight/bias as before

From the docs it sounds like it should be supported to initialize weights/biases using numpy arrays. Any help is appreciated

jaliyae:
Seems that you cannot set the weights directly. [This ](https://stackoverflow.com/questions/43289935/weights-bias-initialization-with-constants-cntk)was a previous answer to the question.
However, since the layer exposes W and b you can assign them values after creating the layer.

FlyinTeller:
To quote the official [docs](https://cntk.ai/pythondocs/cntk.layers.layers.html): 

> init (scalar or NumPy array or cntk.initializer, defaults to glorot_uniform()) – initial value of weights W
> init_bias (scalar or NumPy array or cntk.initializer, defaults to 0) – initial value of weights b
So setting the weights and biases should work. It is even added to the answer of the question you linked. But for some reason it does not seem to be working

[...]

PR #3592, opened 2019-03-10
Fix Dense init bug where initialisation of kernel weights with a numpy array raises a TypeError

delzac:
Allow the initialisation of the kernel weight to be directly initialised by a numpy array, in line with the documentation of Dense layer.

Previously, before this amendment, there is a bug (#2913 ) where initialising kernel weight with numpy array raises
`TypeError: in method 'random_initializer_with_rank', argument 1 of type 'CNTK::ParameterInitializer const &'`

This method of allowing dense kernel weight to be initialised by a numpy array is the same as what you see in the all the Convolution API Layers.


delzac:
@liqunfu @KeDengMS @BowenBao Like to request for review. Its a trivial change, but will allow weights from pre-trained models from other frameworks to be instantiated in cntk.

PR #3723, opened 2019-07-18
remove an assert that seems to be too conservative

yangchen-MS:


liqunfu:
please use VS-DevOp pipeline to merge

PR #3725, opened 2019-07-22
add FutureValue Op's custom attributes into Scan node's description

yangchen-MS:


Source/CNTKv2LibraryDll/proto/onnx/CNTKToONNX.cpp
```diff
@@ -5277,6 +5281,29 @@ bool CNTKToONNXHelper::ProcessLoopsAndCheckCNTKNodeContinueCreate(const Function
...
+                    // FutureValue's custom attrubute will be serialized into the Scan node's description
+                    auto fvOp = scanLoopState.m_stateOutput.Owner();
+                    if (fvOp && fvOp->OpName() == L"FutureValue")
+                    {
+                        const Dictionary& dict = fvOp->GetCustomAttributes();
+                        if (dict.Size() > 0)
+                        {
+                            if (futureValueCustomAttrStr == "")
+                            {
+                                futureValueCustomAttrStr = "{custom_attributes:" + SerializeDictionaryToString(dict) + "}";
```
KeDengMS: json dict keys as string needs to be quoted. Please use "{\"custom_attributes\":"

PR #3854, opened 2021-08-03
CUDA 11.4 from 11.1

nietras:
Many more overloads needed and different hacks to get this compiling.
Many warnings so disabled treat warnings as errors in `MathCUDA`.


nietras:
WRONG REPO

